PR #2, opened 2019-08-02
Feat/extract unit

JarbasAl:
adds nice_unit and, expand_unit from https://github.com/MycroftAI/mycroft-core/pull/1824

adds extract_units


lingua_franca/parse.py
```diff
@@ -334,6 +335,20 @@ def get_gender(word, context="", lang=None):
     return None
 
 
+def extract_units(text, lang=None):
```
JarbasAl: missing docstrings

PR #3, opened 2019-08-02
Feat/huge numbers

JarbasAl:
same as https://github.com/MycroftAI/mycroft-core/pull/1883

- add support for pronouncing infinity
- if number can't be pronounced automatically use scientific notation


krisgesling:
Looks good to me

PR #5, opened 2019-08-02
extract_date_time - 'next {day}' to be 3-9 days in future

JarbasAl:
same as https://github.com/MycroftAI/mycroft-core/pull/2184

    The "next" instance of a day or weekend is considered to be no earlier than
    48 hours in the future. On Friday, "next Monday" would be in 3 days.
    On Saturday, "next Monday" would be in 9 days.

NOTE: this test case is different for some reason, needs investigation https://github.com/MycroftAI/mycroft-core/pull/2184/files#diff-01390ace1c1a3cf68bc710dca7ad532dR359

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -679,6 +679,10 @@ def extract_datetime_en(string, dateNow, default_time):
     date and the remainder string
        "what is weather forecast".
 
+    The "next" instance of a day or weekend is considered to be no earlier than
+    48 hours in the future. On Friday, "next Monday" would be in 3 days.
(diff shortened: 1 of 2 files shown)


krisgesling:
This loop of `enumerate(words)` on [line 764](https://github.com/MycroftAI/lingua-franca/blob/32e6b8e20b200d418e415da316838a3286b02adb/lingua_franca/lang/parse_en.py#L764) seems to be stripping the final day from the utterance and hence not adding a final offset for "next Friday"

I'm assuming something in the `Extras` is overriding an existing method, as the rest of the code is consistent with core.

Will need to come back to this.

krisgesling:
Closing in favour of #60

Issue #6, opened 2019-08-06
Parsing ordinals in form '7th' fails


krisgesling:
Some STT providers return ordinals in the form of digit + suffix eg "7th" rather than "seventh".

Currently extract_number is unable to handle this form regardless of whether the ordinals flag is True or False i.e. the following tests both fail:
```
self.assertEqual(extract_number("this is the 7th test", ordinals=True), 7)
self.assertEqual(extract_number("this is the 7th test", ordinals=False), 7)
```

Ideally this should be handled in the same way as other ordinals to prevent the need to manually extract numbers depending on the STT's output style.

JarbasAl:
this case should work even without the ordinals=True flag, will open a PR for it

JarbasAl:
solved by https://github.com/MycroftAI/lingua-franca/pull/52

PR #9, opened 2019-08-09
handle extracting numbers in the format "5th", "4th", "Nth"

JarbasAl:
solves #6

handles support for explicit ordinal numbers, "1st", "2nd", "3rd", "4th"...."Nth"

independent of ordinals flag


krisgesling:
It's looking good, however there are two more test cases that I think would be worthwhile:
```
self.assertEqual(extract_number("you are the second one", ordinals=True), 2) # PASS
self.assertEqual(extract_number("you are the 8th one", ordinals=True), 8) # FAIL
```

PR #10, opened 2019-08-09
handle extracting "tonight"/"at night"

### LINKED ISSUE #7: extract_datetime("tonight") without explicit time returns only date
`extract_datetime("do something at 9 tonight", now)`
returns the date with time as 9pm correctly however,
`extract_datetime("do something tonight", now)`
returns the date but not a relevant time.

Would expect that if no explicit time (eg '9') is found, that it would revert to a relevant arbitrary value in the same way as [morning=8am, afternoon=3pm, evening=7pm].

Example utterances:
- "what's the weather like tonight"
- "remind me to take the rubbish out tonight"

Perhaps an unspecified tonight should equate to 9pm or later? Or are the use cases too dependent on the Skill?

JarbasAl:
partially fixes #7 

this PR makes the extracted date correct and sets it to 22 pm, but the remainder is imperfect

some later code seems to depend on "tonight" being present in the remainder, some code refactor is in order, documented in unittests

### DIFF
--- a/test/test_parse.py
+++ b/test/test_parse.py
@@ -477,6 +477,13 @@ class TestNormalize(unittest.TestCase):
         testExtract("set alarm at 7:30 on weekdays",
                     "2017-06-27 19:30:00", "set alarm on weekdays")
 
+        # TODO this test is imperfect due to "tonight" in the reminder, but let is pass since the date is correct
+        testExtract("lets meet tonight",
+                    "2017-06-27 22:00:00", "lets meet tonight")
+        # TODO this test is imperfect due to "at night" in the reminder, but let is pass since the date is correct
+        testExtract("lets meet later at night",
+                    "2017-06-27 22:00:00", "lets meet later at night")
+
     def test_extract_ambiguous_time_en(self):
         morning = datetime(2017, 6, 27, 8, 1, 2)
         evening = datetime(2017, 6, 27, 20, 1, 2)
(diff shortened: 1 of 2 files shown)


test/test_parse.py
```diff
@@ -477,6 +477,13 @@ def testExtract(text, expected_date, expected_leftover):
         testExtract("set alarm at 7:30 on weekdays",
                     "2017-06-27 19:30:00", "set alarm on weekdays")
 
+        # TODO this test is imperfect due to "tonight" in the reminder, but let is pass since the date is correct
+        testExtract("lets meet tonight",
+                    "2017-06-27 22:00:00", "lets meet tonight")
+        # TODO this test is imperfect due to "at night" in the reminder, but let is pass since the date is correct
+        testExtract("lets meet later at night",
+                    "2017-06-27 22:00:00", "lets meet later at night")
+
```
krisgesling: Both pass, but may be worth adding:
```
        testExtract("what's the weather like tomorrow night",
                    "2017-06-28 22:00:00", "what is weather like night")
        testExtract("what's the weather like next tuesday night",
                    "2017-07-04 22:00:00", "what is weather like night")
```

krisgesling:
Suggested some extra tests, that are already passing. Will let you decide if you want to add them, but this can be merged with or without them :)

test/test_parse.py
JarbasAl: rebased and additional test added

PR #11, opened 2019-08-12
feat/add nice bytes

JarbasAl:
adds a utility function, `nice_bytes`

converts a number of bytes into a spoken form or unit form, handles all prefixes

```python
nice_bytes(1024) == "1.0 Kilo bytes"
nice_bytes(1024, speech=False) == "1.0 Ki"
```

# TODO
configurable number of decimal places

### DIFF
--- a/lingua_franca/format.py
+++ b/lingua_franca/format.py
@@ -245,25 +245,47 @@ date_time_format = DateTimeFormat(os.path.join(os.path.dirname(__file__),
                                                'res/text'))
 
 
-def nice_bytes(number, lang=None, speech=True):
+def nice_bytes(number, lang=None, speech=True, binary=True, gnu=False):
+    """
+
+    :param number: number of bytes (int)
+    :param lang: lang_code, ignored for now (str)
+    :param speech: spoken form (True) or short units (False)
+    :param binary: 1 kilobyte = 1024 bytes (True) or 1 kilobyte = 1000 bytes (False)
+    :param gnu: Kilo Byte / KiB (False) vs K Byte / KB (True)
+    :return: nice bytes (str)
+    """
     # Attribution: http://stackoverflow.com/a/1094933/2444609
 
     lang_code = get_primary_lang_code(lang)
-    if not speech:
-        default_units = ['B', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
+
+    if speech and gnu:
+        default_units = ['Bytes', 'Kilo', 'Mega', 'Giga', 'Tera', 'Peta', 'Exa', 'Zetta', 'Yotta']
+    elif speech:
+        default_units = ['Bytes', 'Kilo bytes', 'Mega bytes', 'Giga bytes',
+                         'Tera bytes', 'Peta bytes', 'Exa bytes', 'Zetta bytes', 'Yotta bytes']
+    elif gnu:
+        default_units = ['B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
+    elif binary:
+        default_units = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB', 'YiB']
     else:
-        default_units = ['Bytes', 'Kilo bytes', 'Mega bytes', 'Giga bytes', 'Tera bytes',
-                         'Peta bytes', 'Exa bytes', 'Zetta bytes', 'Yotta bytes']
+        default_units = ['B', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi']
+
     #if lang_code == "XX":
     #    units = ['Bytes', 'Kilo bytes', 'Mega bytes', 'Giga bytes', 'Tera bytes',
     #             'Peta bytes', 'Exa bytes', 'Yotta bytes']
     #else:
     units = default_units
 
+    if binary:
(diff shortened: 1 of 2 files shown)


JarbasAl:
relevant for discussion

https://en.wikipedia.org/wiki/Binary_prefix#Specific_units_of_IEC_60027-2_A.2_and_ISO.2FIEC_80000
https://en.wikipedia.org/wiki/Kibibyte
https://stackoverflow.com/questions/1094841/reusable-library-to-get-human-readable-version-of-file-size/

krisgesling:
Not sure if it's just personal preference, but I would default `binary=False`. At the end of the day though, both are available, so I'm happy if no one else minds.

Thanks again for all the bug fixes too!

JarbasAl:
i have been thinking about that, since it's configurable i think the default should be what the majority expects, and i would agree that means binary = False, you do not hear people talking about kibibytes usually, even if that's what they probably mean....

ChanceNCounter:
Looks good at first glance. I just wanna try to break it for a while before closing the review.

JarbasAl:
after the refactor i can give these old PRs a clean up, if you promise to review them this time ;)

JarbasAl:
reopening this one for hacktober fest since it doesnt really impact the ongoing refactoring

lingua_franca/format.py
```diff
@@ -242,7 +242,60 @@ def year_format(self, dt, lang, bc):
...
+    :param places: decimal places (int), TODO not yet implemented
+    :return: nice bytes (str)
+    """
+    lang_code = get_primary_lang_code(lang)
+
+    if speech and gnu:
+        default_units = ['Bytes', 'Kilo', 'Mega', 'Giga', 'Tera', 'Peta', 'Exa', 'Zetta', 'Yotta']
+    elif speech and binary:
+        default_units = ['Bytes', 'Kibibytes', 'Mebibytes', 'Gibibytes',
+                         'Tebibytes', 'Pebibytes', 'Exbibytes', 'Zebibytes', 'Yobibytes']
```
ChanceNCounter, line 270: I think we should check this in a couple different TTS engines, to make sure Mycroft can actually say "kibibytes" and "exbibytes"

lingua_franca/format.py
```diff
@@ -242,7 +242,60 @@ def year_format(self, dt, lang, bc):
...
+    elif gnu:
+        default_units = ['B', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y']
+    elif binary:
+        default_units = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB', 'YiB']
+    else:
+        default_units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+
+    #if lang_code == "XX":
+    #    units = ['...']
+    #else:
```
ChanceNCounter, line 283: Good thought. Less kind thought: should this stuff move to common_data, or perhaps res?

lingua_franca/format.py
```diff
@@ -242,7 +242,60 @@ def year_format(self, dt, lang, bc):
...
+    prefixes - https://en.wikipedia.org/wiki/Binary_prefix
+    spoken binary units - https://en.wikipedia.org/wiki/Kibibyte
+    implementation - http://stackoverflow.com/a/1094933/2444609
+
+    :param number: number of bytes (int)
+    :param lang: lang_code, ignored for now (str)
+    :param speech: spoken form (True) or short units (False)
+    :param binary: 1 kilobyte = 1024 bytes (True) or 1 kilobyte = 1000 bytes (False)
+    :param gnu: say only order of magnitude (bool)  - 100 Kilo (True) or 100 Kilobytes (False)
+    :param places: decimal places (int), TODO not yet implemented
```
ChanceNCounter, line 261: Maybe `if places != 1: raise NotImplementedError`

test/test_format.py
```diff
@@ -419,7 +505,7 @@ def test_nice_year(self):
                 self.assertTrue(len(nice_year(dt, lang=lang)) > 0)
                 # Looking through the date sequence can be helpful
 
-#                print(nice_year(dt, lang=lang))
+    #                print(nice_year(dt, lang=lang))
```
ChanceNCounter, line 508: stray `print()` (not part of this PR, but pylint seems to have exposed it. thanks, pylint!)

ChanceNCounter (changes requested):
Reviewed, though I haven't yet had a chance to check the branch out and try it in the interpreter.

ChanceNCounter:
Incidentally, how universal is this vocabulary? Is this likely to be localized? We could move it straight to English, if so, and let the wrapper whine about missing implementations in other languages, but maybe English is close enough for most purposes.

lingua_franca/format.py
JarbasAl: if a TTS engine cant read that is out of scope from LF, in the end this repo is about formatting and parsing, TTS doesnt even come into the picture until downstream uses it

Mimic2 has a mechanism for users to correct pronunciations, something like that could be expanded for other engines, but this is very implementation specific and out of scope not only of this PR but of this repo

ChanceNCounter: Yeah. I was thinking about this with respect to moving it to common_data, which would be easier for downstream to override via config.

ChanceNCounter:
We shouldn't have let this go stale, reopening

JarbasAl:
Closing, enough time has passed that i dont remember what this code even does

Issue #13, opened 2019-08-13
Bug: extract_duration return phrase may be modified


krisgesling:
In cases where an utterance passed to extract_duration has both a fraction and a duration, the duration is correctly returned, however the fraction in the remaining text string is returned as a float.

This seems to occur only for fraction terms, not ordinals eg "third" but not "second"

Current behaviour:
`extract_duration("the sixth 10 minute timer")[1] == "the 0.166666666666666   timer"`
`extract_duration("the first 10 minute timer")[1] == "the first   timer"`
`extract_duration("the sixth timer for 10 minutes")[1] == "the 0.166666666666666 timer for"`
`extract_duration("the sixth timer for 10 minutes")[1] == "the first timer for"`

Example tests:
self.assertEqual(extract_duration("the first 10 minute timer"), (9000, "the first   timer"))
self.assertEqual(extract_duration("the sixth 10 minute timer"), (9000, "the sixth   timer"))
self.assertEqual(extract_duration("the first timer for 10 minutes"), (9000, "the first timer for"))
self.assertEqual(extract_duration("the sixth timer for 10 minutes"), (9000, "the sixth timer for"))

vegerot:
See #16

JarbasAl:
related to #95 , this adds a replace_token to keep track of consumed words in the remainder, however the number normalization step converts all words to numbers

added a TODO in the unittests

PR #16, opened 2019-08-14
Do not format ordinal numbers.  Addresses #13

vegerot:
Addresses https://github.com/MycroftAI/lingua-franca/issues/13

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -648,14 +648,15 @@ def extract_duration_en(text):
     }
 
     pattern = r"(?P<value>\d+(?:\.?\d+)?)\s+{unit}s?"
-    text = _convert_words_to_numbers(text)
+    number_text = _convert_words_to_numbers(text)
 
     for unit in time_units:
         unit_pattern = pattern.format(unit=unit[:-1])  # remove 's' from unit
-        matches = re.findall(unit_pattern, text)
+        matches = re.findall(unit_pattern, number_text)
         value = sum(map(float, matches))
         time_units[unit] = value
         text = re.sub(unit_pattern, '', text)
+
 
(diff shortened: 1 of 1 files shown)


krisgesling:
Hi Max, 

This is working for the example tests I included, however I did not account for durations that are expressed in words rather than digits eg:
```
self.assertEqual(extract_duration("seven hours"),
                         (timedelta(hours=7), "seven hours"))
self.assertEqual(extract_duration("Four and a half minutes until sunset"),
                         (timedelta(minutes=4.5), "until sunset"))
```
These seem to be failing with the proposed change.

I also had another test fail as the remainder of the utterance is currently returned all lowercase, whilst this update returns the string with the original capitalization. Eg
```
self.assertEqual(extract_duration("Set a timer for 30 minutes"),
                         (timedelta(minutes=30), "set a timer for")) # FAILS
self.assertEqual(extract_duration("Set a timer for 30 minutes"),
                         (timedelta(minutes=30), "Set a timer for")) # PASSES
```
I can't see any reason why the case of the return string would matter, unless anyone else has an opinion? 

We could either intentionally lowercase the return string, or update the tests that fail. Thoughts?

JarbasAl:
i think case should be left as originally, this is the behaviour i would expect, case may convey meaning or even be used in later parsing code (for example comparing to original string could create an unexpected false condition)

we just need to make sure nothing in core depends on this stuff being changed to lower case, i'm not a big fan of fixing tests instead of code! 

as a side note, normalize() will lower case everything, and should usually be called before these methods "in the wild"

krisgesling:
Hey, given how old this PR I'm assuming you're no longer working on it and I'm going to close it. 

Please feel free to re-open if you're keen to tackle the remaining cases.

PR #17, opened 2019-08-16
Refactor/ parse_pt

JarbasAl:
initial refactor of extract_number_pt

- refactor extract_number to share some logic across languages
- default short_scale flag to None, languages can now set a default value
- add portuguese support for short_scale and long_scale
- add portuguese support for ordinals
- add portuguese support for decimals
- extract_number_pt now based on common extract_number logic
- adds extract_numbers_pt
- refactors and improves get_gender_pt
- initial translations of res folder for pt-pt

NOTE: portuguese defaults to long scale


lingua_franca/lang/common_data_pt.py
```diff
@@ -116,3 +133,174 @@
...
+    (1e114, "novendecilião"),
+    (1e120, "vigintilião"),
+    (1e306, "unquinquagintilião"),
+    (1e312, "duoquinquagintilião"),
+    (1e336, "sesquinquagintilião"),
+    (1e366, "unsexagintilião")
+])
+
+_SHORT_SCALE_PT = OrderedDict([
+    (100, 'hundred'),
```
antoniomaciej: You might have missed some Portuguese translations here. Maybe it should be "cem" and "mil" :)

JarbasAl: this is a work in progress, i need to review all strings, i did a simple search and replace to "portuguesify" the names, the prefixes should be the same but i'm sure i missed some things

good catch! :)

antoniomaciej: Sure, I just happened to notice, so I wrote a comment. :)

lingua_franca/lang/parse_pt.py
```diff
@@ -14,26 +14,66 @@
...
+
+_MULTIPLIES_SHORT_SCALE_PT = set(_SHORT_SCALE_PT.values()) | \
+                             generate_plurals_pt(_SHORT_SCALE_PT.values())
+
+_STRING_NUM_PT = invert_dict(_NUM_STRING_PT)
+_STRING_NUM_PT.update(generate_plurals_pt(_STRING_NUM_PT))
+_STRING_NUM_PT.update({
+    "meio": 0.5,
+    "meios": 0.5,
+    "par": 2
```
antoniomaciej: I was also wondering about this for Spanish. Should the text two pairs return 4? If so, should eight dozens return 96? Isn't a pair some sort of object also? For instance, in a hand of poker, a pair should not be translated to 2. What about words like a couple? How do we handle this?

JarbasAl: the reasoning here was that for english "a couple of X" makes sense to be extracted since it is used in everyday language, ("hey mycroft grab a couple of beers")

but in portuguese/spanish there isnt a clear distinction between "pair" and "couple", "a pair of jeans" would not make sense to extract number 2, but "a couple of jeans" would. 

I think we should not consider "pair" == 2, i will remove that from this PR

JarbasAl: in regards to "should eight dozens return 96", i dont think so, this would make sense in [extract units](https://github.com/MycroftAI/lingua-franca/pull/2) since that takes unit dimensions into account

performing math is out of scope for this utility method

antoniomaciej: Thank you for those clarifications, I'll bear them in mind.

adocampo: I think 'pair' and 'couple' are translated into spanish like 'par' and 'pareja'. The former means '2' while the later is used to group things and people in two.

So, I wouldn't remove "pair" == 2...

JarbasAl: let's revisit this topic later, for now i will remove it since i can't think of any relevant example in portuguese where i want to extract it

i left it commented out with a comment pointing to this discussion

lingua_franca/lang/parse_en.py
```diff
@@ -1459,20 +1355,20 @@ def isFractional_en(input_str, short_scale=True):
...
-                fracts[_LONG_ORDINAL_STRING_EN[num]] = num
+                fracts[_LONG_ORDINAL_EN[num]] = num
 
     if input_str.lower() in fracts:
         return 1.0 / fracts[input_str.lower()]
     return False
 
 
-def extract_numbers_en(text, short_scale=True, ordinals=False):
+def extract_numbers_en(text, short_scale=None, ordinals=False):
```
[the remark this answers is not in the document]
JarbasAl: i will need to backtrack to remember why, i remember making this change when adding short_scale=None to extract_numbers to be passed down to the individual language methods, it was necessary for some reason

i think None basically meant to use the default, when passing True or False down the pipeline it was an explicit user choice, i will need to go over the code again to give more details on why this was needed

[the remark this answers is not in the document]
JarbasAl: oh thats it, extract_numbers would either always pass True or False if a default was provided, not allowing a per language default, so a check for None is needed, either in individual methods or in parent extract_numbers method before calling individual languages, doing it in individual_methods keeps language specific code in language specific files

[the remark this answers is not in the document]
krisgesling: Wondering if it will be cleaner to handle this up the chain with something like:
```Python
def extract_numbers(text, short_scale=None, ordinals=False, lang=None):
    if short_scale is None:
        args = (text, ordinals)
    else:
        args = (text, short_scale, ordinals)
    lang_code = get_primary_lang_code(lang)
    if lang_code == "en":
        return extract_numbers_en(*args)
```

lingua_franca/lang/common_data_pt.py
```diff
@@ -116,3 +148,174 @@
...
+    1e18: "trilionésimo",
+    1e24: "quadrilionésimo",
+    1e30: "quintilionésimo",
+    1e36: "sextilionésimo",
+    1e42: "septilionésimo",
+    1e48: "octilionésimo",
+    1e54: "nonilionésimo",
+    1e60: "decilionésimo"
+    # TODO > 1e60
+}
```
adocampo: Shouldn't be here also the femenine forms? It should be modified just like the numbers dictionaries
```
"primeiro": 1,
"primeira": 1,
...
```

JarbasAl: that's a good point, but i think that would make more sense as a replacement while pronouncing, maybe an helper function to switch_gender

that could also be added into the text normalization step for all numbers so they are extracted correctly

seems strange to maintain versions for plural/singular/female/male if it can be done with a rule based approach

lingua_franca/lang/parse_en.py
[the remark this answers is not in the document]
JarbasAl: i like that, will make changes

lingua_franca/lang/common_data_pt.py
```diff
@@ -1,7 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from collections import OrderedDict
+
 # Undefined articles ["um", "uma", "uns", "umas"] can not be supressed,
 # in PT, "um cavalo" means "a horse" or "one horse".
 
-_PT_ARTICLES = ["o", "a", "os", "as"]
+_PT_ARTICLES = {"o", "a", "os", "as"}
```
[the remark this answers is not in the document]
JarbasAl: i agree, that should be done across all language files, in a follow up PR i intend to move this into a .json file based approach for better localization, but for now i will make that change

JarbasAl:
> Also some general feedback, I find it really useful when reviewing to keep refactoring and functionality changes separate commits. For example commit [ad571f6](https://github.com/MycroftAI/lingua-franca/commit/ad571f6afc96263037440e079f14a531b7b88800) both abstracts common functions from English as well as refactoring and adding functionality to the Portuguese.
> 

I agree with that, i wanted to add functionality and the refactor was required half way through so even the commit history got really messy. if it is too much trouble i will close this PR and redo it in smaller ones

lingua_franca/lang/common_data_pt.py
```diff
@@ -116,3 +148,174 @@
...
+    (1e2103, "septingentilião"),
+    (1e2403, "octingentilião"),
+    (1e2703, "nongentilião"),
+    (1e3003, "millinilião")
+])
+
+_ORDINAL_BASE_PT = {
+    1: 'primeiro',
+    2: 'segundo',
+    3: 'terceiro',
```
adocampo: On `_PT_NUMBERS` there are also these ordinals. I guess those should be deleted and let them just here

JarbasAl:
i will split this in smaller PRs, thank you all for the feedback

krisgesling:
Yeah, so easy to do, I constantly have to catch myself doing this too :)
Will checkout the new PR's

PR #29, opened 2019-10-04
improve get_gender PT

JarbasAl:
splitting #17 in smaller PRs

improves logic for get_gender_pt, logic expanded to handle all cases

implemented in a way easy for re use across languages that share the same basic rules


krisgesling:
Thanks Jarbas, looks sane and the tests are passing :)

JarbasAl:
@krisgesling ping

krisgesling:
Thanks for the ping :)

PR #36, opened 2019-10-21
singularize/pluralize

### LINKED ISSUE #35: new utils - singularize and pluralize
i propose two new methods, useful for normalization but also for generating more natural dialog

singularize -> makes word singular
pluralize -> makes word plural form

these would be localized per language

JarbasAl:
closes #35

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 python-dateutil==2.6.0
+inflection
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 python-dateutil==2.6.0
-inflection
+inflection>=0.3.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
-python-dateutil==2.6.0
\ No newline at end of file
+python-dateutil==2.6.0
+inflection
```
ChanceNCounter: I think this should get a version tag, just in case of future bugs or licensing changes.

JarbasAl: this is also a [single file](https://github.com/jpvanhal/inflection/blob/master/inflection.py), would it make sense instead to bundle it in lingua franca? the advantage would be for other languages to use as template

otherwise i agree that we should pin version number

JarbasAl:
to be rebased and integrated with #167

PR #37, opened 2019-10-22
Feat/plural singular pt

JarbasAl:
depends on #36 

portuguese support for #35 

still missing unittests


JarbasAl:
to be rebased and integrated with #167

PR #39, opened 2019-10-22
Bugfix/extract datetime bugs

JarbasAl:
check unittests for edge cases handled

fixes syntax error when trying to access wordPrev first digit but wordPrev is an empty string

fix "after/from this weekday"  ("this" not currently handled) - possibly related to https://github.com/MycroftAI/mycroft-core/issues/2369


forslund:
This looks good and tests are passing. Merging.

PR #42, opened 2019-10-25
universal normalizer

JarbasAl:
implements a base class to handle normalization across all languages

more updates soon


forslund:
Merged through #55. Closing

PR #52, opened 2019-12-11
Bug/extract number en handle "nth"

JarbasAl:
rebase of https://github.com/MycroftAI/lingua-franca/pull/9

support for explicit ordinal numbers, "1st", "2nd", "3rd", "4th"...."Nth"

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -338,6 +338,16 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
             number_words.append(token)
             continue
 
+        # explicit ordinals, 1st, 2nd, 3rd, 4th.... Nth
+        if word.endswith("st") and is_numeric(word[:-2]) and word[:-2].endswith("1"):
+            word = word[:-2]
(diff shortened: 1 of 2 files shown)


lingua_franca/lang/parse_en.py
```diff
@@ -341,6 +341,25 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
...
+
+            # explicit ordinals, 1st, 2nd, 3rd, 4th.... Nth
+            if word.endswith("st") and word[:-2].endswith("1"):
+                word = word[:-2]
+            elif word.endswith("nd") and word[:-2].endswith("2"):
+                word = word[:-2]
+            elif word.endswith("rd") and word[:-2].endswith("3"):
+                word = word[:-2]
+            else:
+                word = word[:-2]
```
ChanceNCounter: I'm a little confused here, though it's possible I'm just tired. These conditions look like they always end the same, no? `word = word[:-2]` in all four cases.

I'm also wondering whether `if word[:-2].endswith('<char>')` could be `if word[-3] is <char>`

JarbasAl: indeed, thats a leftover from a previous approach, guess it was me who was tired!

PR #57, opened 2019-12-13
Add default value to `denominator`

forslund:
Originally submitted to mycroft-core by @ChanceNCounter 

- This affects nice_number and convert_mixed_fraction
- Add tests for convert_mixed_fraction

### DIFF
--- a/lingua_franca/lang/format_common.py
+++ b/lingua_franca/lang/format_common.py
@@ -16,7 +16,7 @@
 #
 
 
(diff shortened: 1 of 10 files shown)


krisgesling:
Looks good.

Is it worth clarifying the accuracy check in the docstring as to why some numbers would return None? 
> Returns the closest fractional representation using the given denominators, provided the result is accurate to 0.01.  For example, 4.500002 would become the whole number 4, the numerator 1 and the denominator 2

PR #61, opened 2019-12-14
split % and # while tokenizing

forslund:
Handles things like 75% and "this is pr #61"

`75%` -> `75 %`
`#61` -> `# 61`
`hashtag #1world` will still be `hashtag #1world`

### DIFF
--- a/lingua_franca/lang/parse_common.py
+++ b/lingua_franca/lang/parse_common.py
@@ -137,7 +137,12 @@ class Normalizer:
         for idx, w in enumerate(words):
             if w in self.stopwords:
                 words[idx] = ""
+        #if words[-1] == '-':
+        #    words = words[:-1]
         utterance = " ".join(words)
+        # Remove trailing whitespaces from utterance along with orphaned
+        # hyphens, more characters may be added later
+        utterance = re.sub(r'- *$', '', utterance)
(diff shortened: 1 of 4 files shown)


JarbasAl:
i like the way you handled this, maybe also add some tests to the tokenize method specifically?

forslund:
Absolutely, will push that later (probably tomorrow)

forslund:
Updated with some basic tests of the tokenize function.

JarbasAl:
looks great, we need to add those to portuguese normalizer (overrides tokenize), want to do it? otherwise i can do it in a follow up PR

forslund:
I'll leave it to you in a followup PR, I wouldn't know how to write a testcase in Portuguese :)

JarbasAl:
the only extra in portuguese is splitting at "-", so "amo-te" becomes ["amo", "te"]

This particular example means "i love you", so its equivalent of splitting it into ["love", "you"]

i can do a follow up PR, but i'm no good with regex so i'm sure you would do it better :p

forslund:
I'll add that and you can write a couple of test cases for it maybe?

forslund:
The PT one is also handled now. Since no new comments I'll merge this now.

PR #62, opened 2019-12-23
Add fixes to spanish by ChanceNCounter

forslund:
This handles parsing cuarta among other smaller issues.

### DIFF
--- a/lingua_franca/lang/parse_es.py
+++ b/lingua_franca/lang/parse_es.py
@@ -120,7 +120,7 @@ def extractnumber_es(text, short_scale=True, ordinals=False):
             break
 
         # number word and fraction
-        ands = ["e"]
+        ands = ["e", "y"]
         if next_word in ands:
             zeros = 0
             if result is None:
--- a/lingua_franca/lang/parse_es.py
+++ b/lingua_franca/lang/parse_es.py
@@ -37,15 +37,14 @@ def isFractional_es(input_str):
     if input_str.endswith('s', -1):
         input_str = input_str[:len(input_str) - 1]  # e.g. "fifths"
 
-    aFrac = ["medio", "media", "tercio", "cuarto", "cuarta", "quinto",
-             "quinta", "sexto", "sexta", "séptimo", "séptima", "octavo",
-             "octava", "noveno", "novena", "décimo", "décima", "onceavo",
-             "onceava", "doceavo", "doceava"]
+    aFrac = {"medio": 2, "media": 2, "tercio": 3, "cuarto": 4,
+             "cuarta": 4, "quinto": 5, "quinta": 5, "sexto": 6, "sexta": 6,
+             "séptimo": 7, "séptima": 7, "octavo": 8, "octava": 8,
+             "noveno": 9, "novena": 9, "décimo": 10, "décima": 10,
+             "onceavo": 11, "onceava": 11, "doceavo": 12, "doceava": 12}
 
     if input_str.lower() in aFrac:
-        return 1.0 / (aFrac.index(input_str) + 2)
-    if (input_str == "cuarto" or input_str == "cuarta"):
-        return 1.0 / 4
+        return 1.0 / aFrac[input_str]
     if (input_str == "vigésimo" or input_str == "vigésima"):
         return 1.0 / 20
     if (input_str == "trigésimo" or input_str == "trigésima"):
@@ -193,6 +192,7 @@ def extractnumber_es(text, short_scale=True, ordinals=False):
     return result
 
 
+# TODO Not parsing 'cero'
 def es_number_parse(words, i):
     def es_cte(i, s):
         if i < len(words) and s == words[i]:
@@ -314,6 +314,7 @@ def normalize_es(text, remove_articles):
     return normalized[1:]  # strip the initial space
 
 
+# TODO MycroftAI/mycroft-core#2348
 def extract_datetime_es(input_str, currentDate=None, default_time=None):
     def clean_string(s):
         # cleans the input string of unneeded punctuation and capitalization
@@ -338,12 +339,12 @@ def extract_datetime_es(input_str, currentDate=None, default_time=None):
             " ").replace(
             "_",
             "")
-        # handle synonims and equivalents, "tomorrow early = tomorrow morning
-        synonims = {"mañana": ["amanecer", "temprano", "muy temprano"],
+        # handle synonyms and equivalents, "tomorrow early = tomorrow morning
+        synonyms = {"mañana": ["amanecer", "temprano", "muy temprano"],
                     "tarde": ["media tarde", "atardecer"],
                     "noche": ["anochecer", "tarde"]}
-        for syn in synonims:
-            for word in synonims[syn]:
+        for syn in synonyms:
+            for word in synonyms[syn]:
                 s = s.replace(" " + word + " ", " " + syn + " ")
         # relevant plurals, cant just extract all s in pt
         wordlist = ["mañanas", "tardes", "noches", "días", "semanas",
--- a/lingua_franca/lang/parse_es.py
+++ b/lingua_franca/lang/parse_es.py
@@ -19,6 +19,7 @@
 """
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
+from dateutil.tz import gettz
 from lingua_franca.lang.format_es import pronounce_number_es
 from lingua_franca.lang.parse_common import *
 from lingua_franca.lang.common_data_es import _ARTICLES_ES, _NUM_STRING_ES
@@ -321,7 +322,7 @@ def extract_datetime_es(input_str, currentDate=None, default_time=None):
         # among other things
         symbols = [".", ",", ";", "?", "!", "º", "ª"]
         noise_words = ["entre", "la", "del", "al", "el", "de",
-                       "por", "para", "una", "cualquier", "a",
+                       "para", "una", "cualquier", "a",
                        "e'", "esta", "este"]
 
         for word in symbols:
@@ -691,7 +692,6 @@ def extract_datetime_es(input_str, currentDate=None, default_time=None):
         if wordNext in months:
             used -= 1
         if used > 0:
-
             if start - 1 > 0 and words[start - 1] in lists:
                 start -= 1
                 used += 1
@@ -765,9 +765,9 @@ def extract_datetime_es(input_str, currentDate=None, default_time=None):
                 hrAbs = 21
             used += 1
         # parse half an hour, quarter hour
-        elif word == "hora" and \
+        elif (word == "hora" and
                 (wordPrev in time_indicators or wordPrevPrev in
-                 time_indicators):
+                 time_indicators)):
             if wordPrev == "media":
                 minOffset = 30
             elif wordPrev == "cuarto":
@@ -1022,8 +1022,11 @@ def extract_datetime_es(input_str, currentDate=None, default_time=None):
             datestr = datestr.replace(monthsShort[idx], en_month)
 
         temp = datetime.strptime(datestr, "%B %d")
+        temp = temp.replace(tzinfo=None)
         if not hasYear:
             temp = temp.replace(year=extractedDate.year)
+            print(gettz(temp.tzname()))
+            print(extractedDate.tzname(), temp.tzname())
             if extractedDate < temp:
                 extractedDate = extractedDate.replace(year=int(currentYear),
                                                       month=int(
--- a/lingua_franca/lang/parse_es.py
+++ b/lingua_franca/lang/parse_es.py
@@ -121,7 +121,7 @@ def extractnumber_es(text, short_scale=True, ordinals=False):
             break
 
         # number word and fraction
-        ands = ["e", "y"]
+        ands = ["y"]
         if next_word in ands:
             zeros = 0
             if result is None:
@@ -269,6 +269,7 @@ def es_number_parse(words, i):
 
     return es_number(i)
 
+
 def extract_numbers_es(text, short_scale=True, ordinals=False):
     """
         Takes in a string and extracts a list of numbers.
@@ -285,7 +286,6 @@ def extract_numbers_es(text, short_scale=True, ordinals=False):
     """
     return extract_numbers_generic(text, pronounce_number_es, extractnumber_es,
                                    short_scale=short_scale, ordinals=ordinals)
-
 
 
 def normalize_es(text, remove_articles):
--- a/test/test_parse_es.py
+++ b/test/test_parse_es.py
@@ -99,8 +99,10 @@ class TestNormalize(unittest.TestCase):
 
     def test_isFraction_es(self):
         self.assertEqual(isFractional_es("vigésimo"), 1.0 / 20)
+        self.assertEqual(isFractional_es("vigésima"), 1.0 / 20)
         self.assertEqual(isFractional_es("trigésimo"), 1.0 / 30)
         self.assertEqual(isFractional_es("centésima"), 1.0 / 100)
+        self.assertEqual(isFractional_es("centésimo"), 1.0 / 100)
         self.assertEqual(isFractional_es("milésima"), 1.0 / 1000)
 
 
@@ -110,8 +112,8 @@ class TestDatetime_es(unittest.TestCase):
         # test currentDate==None
         _now = datetime.now()
         relative_year = _now.year if _now.day == 1 else (_now.year + 1)
-        self.assertEqual(extract_datetime_es("11 ene")[0],
-                         datetime(relative_year, 1, 11))
+        #self.assertEqual(extract_datetime_es("11 ene")[0],
+        #                 datetime(relative_year, 1, 11))
 
         # test months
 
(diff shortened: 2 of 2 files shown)


lingua_franca/lang/parse_es.py
```diff
@@ -338,12 +339,12 @@ def clean_string(s):
             " ").replace(
             "_",
             "")
-        # handle synonims and equivalents, "tomorrow early = tomorrow morning
-        synonims = {"mañana": ["amanecer", "temprano", "muy temprano"],
+        # handle synonryms and equivalents, "tomorrow early = tomorrow morning
```
ChanceNCounter: I seem to have replaced a typo in this comment with a different typo

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+        self.assertEqual(extract_number("dos e media", lang='es'), 2.5)
+        self.assertEqual(extract_number(
+            "catorce e milésima", lang='es'), 14.001)
+
+        self.assertEqual(extract_number("dos punto cero dos", lang='es'), 2.02)
+
+    def test_isFraction_es(self):
+        self.assertEqual(isFractional_es("vigésimo"), 1.0 / 20)
+        self.assertEqual(isFractional_es("trigésimo"), 1.0 / 30)
+        self.assertEqual(isFractional_es("centésima"), 1.0 / 100)
```
adocampo: Also the male form can be said: "centésimo"

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
         self.assertEqual(extract_number("seis punto dos", lang='es'), 6.2)
         self.assertEqual(extract_numbers("un medio", lang='es'), [0.5])
+        self.assertEqual(extract_number("cuarto", lang='es'), 0.25)
+
+        self.assertEqual(extract_number("2.0", lang='es'), 2.0)
+        self.assertEqual(extract_number("1/4", lang='es'), 0.25)
+
+        self.assertEqual(extract_number("dos e media", lang='es'), 2.5)
+        self.assertEqual(extract_number(
+            "catorce e milésima", lang='es'), 14.001)
```
adocampo: In spanish we would say: "catorce coma cero cero uno" I hardly doubt someone says "catorce y una milésima"

lingua_franca/lang/parse_es.py
forslund: Thanks, updated the commit.

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+        self.assertEqual(extract_datetime(
+            "11 dic", lang='es', anchorDate=datetime(1998, 12, 1))[0],
+            datetime(1998, 12, 11))
+
+        self.assertEqual(extract_datetime("", lang='es'), None)
+
+    # TODO fix bug causing these tests to fail (MycroftAI/mycroft-core#2348)
+    #         reparar error de traducción preveniendo las funciones abajo de
+    #         retornar correctamente
+    #         (escribido con disculpas por un Inglés hablante)
```
adocampo: escrito  ;)

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+        # It's also failing on years
+        self.assertEqual(extract_datetime(
+            "11 ago 1998", lang='es')[0], datetime(1998, 8, 11))
+
+    def test_extract_datetime_relative(self):
+        self.assertEqual(extract_datetime("esta noche", anchorDate=datetime(
+            1998, 1, 1), lang='es'),
+            [datetime(1998, 1, 1, 21, 0, 0), 'esta'])
+        self.assertEqual(extract_datetime(
+            "ayer noche", anchorDate=datetime(1998, 1, 1),
```
adocampo: Yet "ayer por la noche" or "ayer noche" aren't wrong at all, we rather use "anoche" to specify "last night".

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+            lang='es')[0], datetime(1997, 12, 31, 21))
+        self.assertEqual(extract_datetime(
+            "el noche anteayer", anchorDate=datetime(1998, 1, 1),
```
adocampo: We say "hace dos noches" or "anteanoche"

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+            lang='es')[0], datetime(1997, 12, 30, 21))
+        self.assertEqual(extract_datetime(
+            "el noche ante ante ayer", anchorDate=datetime(1998, 1, 1),
```
adocampo: We say "hace tres noches" ("anteanteanoche" and "trasanteanoche" are no longer on the spanish dictionary since 2008 because no one say that way)

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+            lang='es')[0], datetime(1997, 12, 29, 21))
+        self.assertEqual(extract_datetime(
+            "mañana en la mañana", anchorDate=datetime(1998, 1, 1),
```
adocampo: In proper spanish or es_ES, should be said "mañana **por** la mañana". _Some_ latam say "mañana **en** la mañana" but those usually are es_US or es_MX, more influenced by english language.

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+            lang='es')[0], datetime(1998, 1, 2, 8))
+        self.assertEqual(extract_datetime(
+            "ayer en la tarde", anchorDate=datetime(1998, 1, 1),
```
adocampo: same than before. "Mañana **por** la tarde"

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+            lang='es')[0], datetime(1997, 12, 31, 15))
+
+        self.assertEqual(extract_datetime(
+            "que año es", anchorDate=datetime(1998, 1, 1),
```
adocampo: Should it be accented? "qué año es" if it is a question, it should. (I don't know if that makes any sense of the unit test)

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+        self.assertEqual(extract_datetime("hoy 2 am", lang='es',
+                                          anchorDate=datetime(1998, 1, 1))[0],
+                         datetime(1998, 1, 1, 2))
+        self.assertEqual(extract_datetime("hoy 2 p.m.", lang='es',
+                                          anchorDate=datetime(1998, 1, 1))[0],
+                         datetime(1998, 1, 1, 14))
+        self.assertEqual(extract_datetime("hoy 2 pm", lang='es',
+                                          anchorDate=datetime(1998, 1, 1))[0],
+                         datetime(1998, 1, 1, 14))
+
```
adocampo: In spanish we don't say "am" or "pm" but "de la madrugada", "de la mañana", "de la tarde", or "de la noche". The time frames varies from region to region, but usually are those:
- from 00:01 to 11:59 --> "de la mañana"
- from 12:01 to 20:00 --> "de la tarde"
- from 20:01 to 23:59 --> "de la noche"

There are some special time frames:
- 00:00 is "media noche" or "medianoche" and 12:00 is "medio día" or "mediodía"
- the time frame from 01:00 to 05:59 is usually also called "de la madrugada"

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+    @unittest.skip("These phrases are not parsing correctly.")
+    # parses as "morning" and returns 8:00 on anchorDate
+    def test_extract_datetime_relative_failing(self):
+        self.assertEqual(extract_datetime(
+            "mañana", anchorDate=datetime(1998, 1, 1), lang='es')[0],
+            datetime(1998, 1, 2))
```
adocampo: When I tried to create parser_es.py I tried to differentiate "morning" from "tomorrow" by parsing the former word. In spanish "la mañana" is _always_ the time frame before midday, while the day after today is simply "mañana". That's the only way we could differentiate those two terms: if "mañana" is preceded by "la" is unmistakably "morning" and without is unmistakably "tomorrow", so even "mañana por la mañana" can be parsed correctly.

lingua_franca/lang/parse_es.py
```diff
@@ -37,15 +37,14 @@ def isFractional_es(input_str):
     if input_str.endswith('s', -1):
         input_str = input_str[:len(input_str) - 1]  # e.g. "fifths"
 
-    aFrac = ["medio", "media", "tercio", "cuarto", "cuarta", "quinto",
-             "quinta", "sexto", "sexta", "séptimo", "séptima", "octavo",
-             "octava", "noveno", "novena", "décimo", "décima", "onceavo",
-             "onceava", "doceavo", "doceava"]
+    aFrac = {"medio": 2, "media": 2, "tercio": 3, "cuarto": 4,
+             "cuarta": 4, "quinto": 5, "quinta": 5, "sexto": 6, "sexta": 6,
+             u"séptimo": 7, u"séptima": 7, "octavo": 8, "octava": 8,
```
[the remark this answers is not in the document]
forslund: Correct you are! Will remove those

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
             lang='es')), [1, 4, 7, 8, 14, 157])
         self.assertEqual(extract_number("seis punto dos", lang='es'), 6.2)
         self.assertEqual(extract_numbers("un medio", lang='es'), [0.5])
+        self.assertEqual(extract_number("cuarto", lang='es'), 0.25)
+
+        self.assertEqual(extract_number("2.0", lang='es'), 2.0)
+        self.assertEqual(extract_number("1/4", lang='es'), 0.25)
+
+        self.assertEqual(extract_number("dos e media", lang='es'), 2.5)
```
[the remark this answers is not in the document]
forslund: Ah humm, that makes the test fail...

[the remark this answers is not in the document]
forslund: Ok think I fixed that...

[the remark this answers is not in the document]
adocampo: oh, it can also be said "dos y medio", depending if the subject is male or female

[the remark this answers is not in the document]
adocampo: or "dos coma cinco" | "dos punto cinco" if we are speaking of just decimal numbers

lingua_franca/lang/parse_es.py
```diff
@@ -120,7 +120,7 @@ def extractnumber_es(text, short_scale=True, ordinals=False):
             break
 
         # number word and fraction
-        ands = ["e"]
+        ands = ["e", "y"]
```
adocampo: "e" as number or fractional separator doesn't exist in spanish. That's copy pasted from PT or IT. 
"e" can be used as a "y" only and if only the next word begins with an "i", to avoid cacophonies.

forslund: So it can't be used with any numbers?

adocampo: there is no number starting with "i", so no.

forslund: Thanks I'll remove the "e"

forslund: Just for clarity: the test phrace "catorce e milésima" should be "catorce y milésima"

adocampo: TL;DR: yes, but for that example in particular, I also commented this https://github.com/MycroftAI/lingua-franca/pull/62#discussion_r360832559.

But yes, you're right, "**and**" in spanish is "**y**". Always... on numbering. 
It is always "y" when naming numbers and fractions. "Y" is _always_ placed between tens and units. You have some exceptions, like numbers from 11 to 29

45: cuarenta y cinco 
1492: mil cuatrocientos noventa y dos
2.856.729: dos millones ochocientos cincuenta y seis mil setecientos veintinueve
11: once
25: veinticinco

You can find some exceptions finding an "e" when enumerating and the forthcoming word begins with an "i", for example: "Eres guapo e inteligente" (You are handsome and intelligent), as the phonema of "y" and "i" is the same, so we put an "e" to avoid the cacophony.

Returning to numbers, as no number begins with an "i" (uno, dos tres, cuatro, cinco, seis, siete, ocho, nueve, diez, cien, mil, millon, etc), you will never ever see an "e".

forslund: Thanks for the explanation! :+1:

test/test_parse_es.py
ChanceNCounter: I thought this would be a trivial addition, add another condition there, but it turns out there are a number of places in the parser that require some tweaking to facilitate this. If one of you guys wants to take a stab, please do!

test/test_parse_es.py
ChanceNCounter: well, i guess that's the most accurate comment I've ever written, then!

(Ake: comment reads, roughly, "fix translation error preventing below functions from returning correctly. written with apologies by an English speaker." except 'escribido' should read 'escrito')

test/test_parse_es.py
ChanceNCounter: Sorry, I keep forgetting to check all of these. So far, this is the only one where I don't defer to you ;)

This particular test was to ensure that "ayer" and "noche" combined gets yesterday and evening, rather than returning either tonight or yesterday morning, as you'd get if you passed it "ayer" or "noche" separately.

That said, if there isn't one, there should be a test for "anoche" That's an oversight on my part.

test/test_parse_es.py
forslund: This has been updated now.

test/test_parse_es.py
forslund: Adding a test for _anoche_ (to work exactly like ayer noche) does not pass. Let's create an issue for this after the PR is merged.

forslund:
I don't think I can handle all the requested changes. My suggestion is that we merge this in the current state so it's at the same state as the current core. and we add @adocampo as issues allowing people more skilled in Spanish to move those along.

If there's some changes that are too weird I should be able to remove those.

adocampo:
I agree we could merge and start working piece by piece

ChanceNCounter:
Currently working a little bit, since it hasn't been merged yet anyway, and there are plenty of changes listed here to be addressed.

test/test_parse_es.py
```diff
@@ -84,6 +86,134 @@ def test_extract_number_es(self):
...
+        self.assertEqual(extract_number(
+            "catorce e milésima", lang='es'), 14.001)
+
+        self.assertEqual(extract_number("dos punto cero dos", lang='es'), 2.02)
```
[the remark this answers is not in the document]
ChanceNCounter: The parser already handled both, but there weren't any tests for "coma." There are about to be.

test/test_parse_es.py
ChanceNCounter: Agreed. This test is just to make sure that it 1) catches milésima, and 2) correctly combines it with another number. You'd hardly ever say "14 and a thousandth" in English, either, but edge cases are half the point.

That's also what these 'e's were for - cover edge cases - but they're obviously superfluous now that 'e' has been removed. I'd never noticed there aren't any numbers ending in 'i'. I just dutifully made sure both ands were functional.

test/test_parse_es.py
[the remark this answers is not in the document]
ChanceNCounter: I stand corrected, and it's a very good thing you noticed this, and yet another reminder why it's so important to test every possible category of input!

The word "coma" works, but the character `,` is not currently parsing.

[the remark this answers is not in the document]
ChanceNCounter: Apparently nobody implemented logic for written decimals. It's just been passing the whole float through the whole parser. This is now its own issue, I reckon.

forslund:
@ChanceNCounter do you mean you want us to hold off on merging?

ChanceNCounter:
Yes, sorry, I'd have been clearer if I'd realized it was other-people-are-awake o'clock. I'll have this whittled down to "fixed stuff" and "stuff that needs its own PR" in a short while.

test/test_parse_es.py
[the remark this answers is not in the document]
adocampo: Well, I guess we can live just with "punto" until someone can parse the comma. The point here is the human could say "dos coma cero dos" and Mycroft should understand it as "2.04". As Mycroft is a voice assistant, we won't type "2,02" in the near future.

PS: I'm wondering how french or german people deals with this, as afaik, in those languages decimal separator is comma as well.

test/test_parse_es.py
ChanceNCounter: relevant to #64

@adocampo apart from "esta mañana," are there any other exceptions that we'll need to account for?

test/test_parse_es.py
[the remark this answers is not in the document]
ChanceNCounter: Yeah, I've put up an issue to add that logic. Happy to tackle it myself, as well, but I've got appointments through the middle of the day, so I won't get to it until this evening at the earliest.

At its most basic, it's just a regex search for a whole word consisting entirely of numbers separated by a comma.

That won't work in the very long term, because somebody might type in `2,02E`, but it would at least solve `2,02`.

I agree that it's a low priority, as the module is mainly for Mycroft and Mycroft is a voice assistant, but it's entirely conceivable that somebody could - right this minute! - try to parse a bunch of text for Mycroft to pronounce, and it would fail on commas-as-decimals.

test/test_parse_es.py
adocampo: When you say other exceptions you mean all combinations of  "_whatever_ mañana" meaning "morning" don't you? For example
"esta mañana" means "this morning"
"hoy por/en la mañana" means "today in the morning"
"mañana por/en la mañana" means "tomorrow morning"

So, no. I think generally we can say "la mañana" for the time frame between midnight and noon with the sole exception of "esta mañana" which means "this morning".

Good point!

forslund:
Thanks @ChanceNCounter, merging now.

PR #69, opened 2020-01-11
Add logic to normalize comma-delimited decimals

### LINKED ISSUE #65: extract_number() functions need to convert '#,#' to float('#.#')

ChanceNCounter:
Closes #65

~~**All** languages, including English, will now normalize `34,5` to `34.5` before beginning to extract numbers.~~

Decimal markers can now be specified through `extract_number()` and `extract_numbers()` function calls, using a new parameter `decimal='.'`

Note that these functions will only normalize decimals if they are called *as such*. Individual parsers, such as `extractnumber_es()`, have not been modified in any way, but will produce correct output when called via `extract_number(lang='es', decimal=',')`

For those who don't speak regex, though I encourage you to run the regex through a regex tester, it means this:

`\b\d+,{1}\d+\b`:
`\b` a word boundary
`\d+` any number of digits, followed by
`,{1}` exactly one comma, followed by
`\d+` any number of digits
`\b` another word boundary

### DIFF
--- a/lingua_franca/parse.py
+++ b/lingua_franca/parse.py
@@ -80,17 +80,20 @@ def match_one(query, choices):
         return best
 
 
-def normalize_decimals(text):
+def normalize_decimals(text, decimal):
     """
-        Replace decimal commas with decimal periods so Python can floatify them
+        Replace 'decimal' with decimal periods so Python can floatify them
     """
-    sanitize_decimals = re.compile(r"\b\d+,{1}\d+\b")
+    regex = r"\b\d+" + decimal + r"{1}\d+\b"
+    sanitize_decimals = re.compile(regex)
     for _, match in enumerate(re.finditer(sanitize_decimals, text)):
-        text = text.replace(match.group(0), match.group(0).replace(',', '.'))
+        text = text.replace(match.group(
+            0), match.group(0).replace(decimal, '.'))
     return text
 
 
-def extract_numbers(text, short_scale=True, ordinals=False, lang=None):
+def extract_numbers(text, short_scale=True, ordinals=False, lang=None,
+                    decimal='.'):
     """
         Takes in a string and extracts a list of numbers.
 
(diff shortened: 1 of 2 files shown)


ChanceNCounter:
run test

ChanceNCounter:
A few minutes ago, I realized the way I was trying to distinguish comma-delimited decimals from comma-delimited thousands wasn't gonna work.

So I came here to mark this WIP, and I now see that commit didn't make it to the branch in any case 😳

ChanceNCounter:
It occurs to me that this argument will almost always be used to parse `','` as a decimal point, and that users who want `','` for decimal points will almost always want `'.'` for thousands separators.

Should this PR address that?

If so, should it be an additional keyword, `thousands=','`, or should the `decimal` keyword be scrapped in favor of actual keywords?

Although the function signatures may become bloated, I'm partial toward two keywords. Standards exist other than full stops or commas for decimal separators. Indeed, it *might* be worth handling spaces, if specifically indicated in the function call, but that's another layer of complexity, probably involving a while loop.

JarbasAl:
in a computer a decimal number is always represented with a .  regardless of language. 

I don't think we will get any stt transcription ever where this isn't the case

ChanceNCounter:
> in a computer a decimal number is always represented with a . regardless of language.
> 
> I don't think we will get any stt transcription ever where this isn't the case
> 
> It's still good to think about this, chat usage is also important !

Exactly. This PR is so that people can parse data written the opposite way. @thelastproject asked and shall receive.

ChanceNCounter:
Alright, here we go, now that I'm free and at a desktop. Say you wanted TTS to read lines from a document, containing yearly *something*, and you knew the file's formatting: `YYYY: #,#`

With this PR, you could do:

```python
>>> foo = parse.extract_numbers("1942: 4,5".replace(":",""), decimal=",")
>>> foo
[1942.0, 4.5]
>>> d = format.nice_year(datetime(year=int(foo[0]), month=1, day=1))
>>> d
'nineteen forty two'
>>> bar = d + ", " + format.pronounce_number(foo[1])
>>> bar
'nineteen forty two, four point five'
```

Pass that along to TTS and you've got some nice behavior.

PR #76, opened 2020-01-27
fix #75 pronounce_number fails on num < 1.0

### LINKED ISSUE #75: Fails with 0.0x numbers
```
$ python3
Python 3.7.6 (default, Dec 19 2019, 22:52:49) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
@>>> from lingua_franca.format import pronounce_number
@>>> pronounce_number(0.05)
' point'
@>>> pronounce_number(0.05, places=1)
' point'
@>>> pronounce_number(0.05, places=0)
''
```

This causes trouble in the [travel-time-skill](https://github.com/TheLastProject/travel-time-skill) for short distances:
```
        self.speak_dialog(
            dialog,
            {
                'time': time,
                'distance': util.format.pronounce_number(distance, places=0),
                'from': from_,
                'destination': destination
            }
        )
```

Given `distance = 0.05` and the following dialog options:
```
It is {{distance}} kilometers and {{time}} from {{from}} to {{destination}} by car
```

Output will be as follows:
```
It is kilometers and under a minute from A to B by car
```

ChanceNCounter:
Alter format.pronounce_number_en() to parse the decimal part using strings and characters rather than math.

Bug was due to int comparisons clobbering small numbers.

Closes #75

### DIFF
--- a/lingua_franca/lang/format_en.py
+++ b/lingua_franca/lang/format_en.py
@@ -290,8 +290,8 @@ def pronounce_number_en(num, places=2, short_scale=True, scientific=False,
         return pronounce_number_en(num, places, short_scale, scientific=True)
     # Deal with fractional part
     elif not num == int(num) and places > 0:
-        if num < 1.0 and not result:
-            result = "zero"
+        if abs(num) < 1.0 and (result is "minus " or not result):
+            result += "zero"
         result += " point"
         _num_str = str(num)
         _num_str = _num_str.split(".")[1][0:places]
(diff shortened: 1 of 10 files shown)


ChanceNCounter:
Regarding scope: the leading "zero" was added as the least evil of three.

`assertEqual(pronounce_number(0.05), "point zero five")` failed (in Italian, actually) because the function was returning `" point zero five"`.

I could write the test to account for it, but then, when somebody fixed the problem, the test would suddenly fail. I could alter the code simply to avoid adding the space if result is empty, but now I'm making a change, so, why? The other five parsers were already doing "zero point zero five," so I went with standardizing them. Now it's possible to write syntactically correct tests for both definitions of "syntax" :relieved:

JarbasAl:
:+1: i approve of this PR

i think the leading zero makes sense, and it is correct, "point zero five" is basically a verbal shortcut because humans are lazy

ChanceNCounter:
Just discovered the leading zero doesn't work on negative numbers.

ChanceNCounter:
Think that oughta do it.

PR #71, opened 2020-01-20
fix: extractnumber_xx API incompatibility

YuukanOO:
For some languages, the definition of `extractnumber_xx` was incompatible with the use of `extract_numbers_generic` since it lacks `short_scale` and `ordinals` keyword args.

This PR fix this.

It also updates the `extractnumber_fr` when no number has been found on the given text by returning `False` instead of a normalized string to make it in sync with all other `extractnumber_xx` methods in other languages which return a number or False if no number has been parsed. Doing so handle cases such as `extractnumbers('2 heures')` which failed otherwise.

PR #77, opened 2020-01-27
Fix/extractnumber incompatibility

ChanceNCounter:
Merges #71 to master, with the syntax fix ported to the Spanish and Portuguese parsers.


ChanceNCounter:
Thanks again to @YuukanOO!

Issue #83, opened 2020-02-23
Hyphens are not handled


JeremyStepp:
By many it is considered proper to write numbers like "sixty seven" as "sixty-seven". Currently, hyphens make the numbers invalid. It seems like simply replacing hyphens with spaces before tokenization would solve the issue. I know that hyphens are also used in French.

ChanceNCounter:
It would be important to avoid removing *leading* hyphens, as this would mess with negative numerals (-2.5)

JarbasAl:
would this make sense to be handled in the [normalizer](https://github.com/MycroftAI/lingua-franca/pull/55)?

ChanceNCounter:
I think we should assume that hyphens will become syntactically important to other parsers. I don't think they'll ever be meaningful for the number extractors, but I suspect their grammatical significance will come up elsewhere.

`extract_number()`-specific normalization does seem to be a growing concern, though. Should we write a "sub-normalizer" for the number extractors?

JarbasAl:
i don't think a subnormalizer makes much sense, just adds extra places to keep track off. 

We already do [words to digits in the Normalizer](https://github.com/MycroftAI/lingua-franca/blob/master/lingua_franca/lang/parse_common.py#L118), so it makes sense to handle this there, and this could even be considered a bug in words to digits

JarbasAl:
actually, i think normalize should be called before extract_numbers which is not done at all

ChanceNCounter:
Yes, but what if some other parser cares about the hyphen? For instance, many French place names have a hyphen in them, which we wouldn't want the normalizer to lose.

On the other hand, this will be the second thing in as many days that the number parsers have to normalize themselves: numbers (and only numbers) need the input converted to lowercase before parsing, and numbers (and only numbers) need hyphens normalized away before parsing.

ChanceNCounter:
Indeed, words to digits will almost certainly break extract_numbers, as it would turn "twenty two" into [20, 2]

ChanceNCounter:
I've been thinking about this, and I'm now thinking subnormalizers might *reduce* each parser's individual complexity, and actually make it easier to track things.

For instance, normalizing a string for `extract_datetime()` could replace "September" with something like "{month:9}" to simplify the crawling portion of the extractor, but other parsers wouldn't want that, because there are other contexts in which the word September *isn't* part of a datetime.

For `extract_number()`, we'd want to remove hyphens *unless* they are found in the form:

[...]

PR #85, opened 2020-02-23
fix #80 - capital letters break extract_number

### LINKED ISSUE #80: Capitals break number parsing
"Ninety four" returns "4" unlike "ninety four" which returns "94". It doesn't recognize the number when the capital is anywhere else in a word as well.

ChanceNCounter:
`extract_number()` helpers (`extractnumber_en()` and so forth):
  - convert input string to lowercase before parsing
  - bug not present in `extractnumber_fr()`

-----

Same/similar change to all but one parser. This could be handled as a
one-liner in the top-level functions, but somebody might import the
lang-specific helpers someday, and this bug would be a nasty surprise.

closes #80

### DIFF
--- a/test/test_parse.py
+++ b/test/test_parse.py
@@ -88,6 +88,13 @@ class TestNormalize(unittest.TestCase):
         self.assertEqual(extract_number("three quarter cups"), 3.0 / 4.0)
         self.assertEqual(extract_number("three quarters cups"), 3.0 / 4.0)
         self.assertEqual(extract_number("twenty two"), 22)
+        self.assertEqual(extract_number(
+            "Twenty two with a leading capital letter"), 22)
+        self.assertEqual(extract_number(
+            "twenty Two with two capital letters"), 22)
+        self.assertEqual(extract_number(
+            "twenty Two with mixed capital letters"), 22)
+        self.assertEqual(extract_number("Twenty two and Three Fifths"), 22.6)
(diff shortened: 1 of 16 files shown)


test/test_parse.py
```diff
@@ -88,6 +88,13 @@ def test_extract_number(self):
         self.assertEqual(extract_number("three quarter cups"), 3.0 / 4.0)
         self.assertEqual(extract_number("three quarters cups"), 3.0 / 4.0)
         self.assertEqual(extract_number("twenty two"), 22)
+        self.assertEqual(extract_number(
+            "Twenty two with a leading capital letter"), 22)
+        self.assertEqual(extract_number(
+            "twenty Two with two capital letters"), 22)
```
JarbasAl: i think you meant to write Twenty Two in here

JarbasAl:
this is simple enough, i think it can be merged, i left a small inconsequential nitpick

test/test_parse.py
ChanceNCounter: Good catch, though I was actually going for, "twenty Two with Two..." I'll fix that.

ChanceNCounter:
Hold pending discussion on #83

ChanceNCounter:
It's been a couple weeks, and no change yet on the normalizer front. I don't think there's any particular reason to keep this PR on hold.

Issue #86, opened 2020-02-23
extractnumber_en() summing "leading hundreds"


ChanceNCounter:
```python
>>> extract_number("ninety nine million")
99000000
>>> extract_number("nine hundred ninety nine million")
99000900

PR #87, opened 2020-03-10
fix #81

### LINKED ISSUE #81: Multiples of ten describing other multiples of ten require an intermediary number. I.e "twenty thousand" doesn't work
"twenty thousand" returns "20". The only way I have found to get this sort of situation to work is to add an extra number. "twenty four thousand" properly returns "24000".

ChanceNCounter:
Fix: multiples of ten which are not powers of ten will no longer fail
to multiply with members of `parse_en.multiplies`.

Closes #81 

I should point out that #86 is still breaking the same function, so this *probably* isn't the whole thing. But as soon as we've got a hotfix for that other one, I recommend publishing a minor version immediately.

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -444,7 +444,11 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 current_val = val
 
         else:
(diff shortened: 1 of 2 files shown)


forslund:
Looks good, merging.

PR #90, opened 2020-03-13
Fix/#86 leading hundreds

### LINKED ISSUE #86: extractnumber_en() summing "leading hundreds"

ChanceNCounter:
closes #86 

Explanation is in the commit message. I suggest using your debugger to fully understand this situation. Also, if anybody else can think of more test cases, that would be ideal.

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -470,16 +470,14 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 # (above, when `current_val` is 1 million, `val` is
                 # 2 million.)
                 #
-                #
-                # We have a dict `string_num_scale` containing [word, value]
-                # for "all" powers of ten.
+                # We have a dict `string_num_scale` containing [value, word]
+                # pairs for "all" powers of ten: string_num_scale[10] == "ten.
                 #
                 # We need go over the rest of the tokens, looking for other
                 # powers of ten. If we find one, we compare it with the current
                 # value, to see if it's smaller than the current power of ten.
                 #
-                # Numbers which are not powers of ten will be passed over,
-                # passing True to the if statement's all() condition.
+                # Numbers which are not powers of ten will be passed over.
                 #
                 # If all the remaining powers of ten are smaller than our
                 # current value, we can set the current value aside for later,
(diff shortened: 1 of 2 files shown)


lingua_franca/lang/parse_en.py
```diff
@@ -460,7 +461,12 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
             # six hundred sixty six
             # two million five hundred thousand
             if word in multiplies and next_word not in multiplies:
-                if all([string_num_scale.get(other_token.word) < current_val if other_token.word in multiplies else True for other_token in tokens[idx::]]):
+                # iterate over the remaining words, and check for larger powers
+                # of ten. if we find one, skip this part, preserving the current
+                # value for the next iteration
+                if all([string_num_scale.get(other_token.word) < current_val
+                        if other_token.word in multiplies else True
+                        for other_token in tokens[idx+1:]]):
```
forslund: The ternary expression tripped me up for a while here :) maybe `or other_token.word not in multiplies` could be considered?

The ternary expression inside the list comprehension made the logic tricky to grasp for me.

Also is the `.get` adding anything here? if the word isn't in `string_num_scale` it will return `None` and the comparison to `current_val` would fail.

ChanceNCounter: `multiplies` and `string_num_scale` derive from the same collection. `multiplies` should be `string_num_scale.values`.

In this case, all the `get()` was really changing was the exception it'd raise, in case we break one of these functions. `get()` would raise a TypeError, index lookup a KeyError. On reflection, the KeyError's clearer, so, good call.

I like the `all()` call because it's an arbitrary number of comparisons. If it *is* time to append
`val` to `to_sum`, we're going to iterate over the entire string from the current word forward. `else True` allows it to work without additional comparisons, while ensuring that we'll stop iterating as soon as we hit one `False`.

OTOH, I recognize that this is confusing, so I've written a VERY long inline explanation.

ChanceNCounter: I suppose it could be a filter, that might be more readable.

ChanceNCounter: Turns out it's better as a for loop.

ChanceNCounter:
If you wind up doing it, rather than I, the merge commit should reflect that the breakpoint lines are no longer anywhere close to where that commit message said they'd be.

lingua_franca/lang/parse_en.py
forslund: I didn't have an issue with the all(), I thought that was quite clear, it was the ternary expression in the list comprehension that tripped me up and I thought an `or` might be clearer. As I said it was a personal preference

ChanceNCounter: I figured, but after staring at it some more, I realized that a straight for loop is much, *much* more efficient here, as the all() arrangement still iterated over the whole string, every time around. This loop stops as soon as it hits False, like all() usually would.

forslund: Ah, good call.

forslund:
Ok, Merging this now. Thanks @ChanceNCounter, sorry for being dense :)

PR #91, opened 2020-03-16
fix #88 and #93: multiple decimal places and multiple decimal numbers

### LINKED ISSUE #88: extract_number does not directly handle multiple decimal places
Calling attention to [this](https://github.com/MycroftAI/lingua-franca/blob/e6837eb2f8fabb72af3a8389c4a85cbcdae5e40b/lingua_franca/lang/parse_en.py#L298-301) TODO.

I'm marking this as a bug because I think that's how users will perceive it, although one could argue that it's just unimplemented functionality.

ChanceNCounter:
* extract_number(), extract_numbers(), and all
   helper functions gain a keyword parameter
   `decimal_places` (for helpers, just `places`)
   which does what it sounds like, using builtin
   round().

 * avoid capturing non-adjacent numbers as decimal
   places

 * avoid capturing already-used decimal places as
   separate numbers in extract_numbers()

 * add a few tests for the above

----

closes #88, #93

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -15,7 +15,6 @@
 #
 from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
-from math import ceil, floor
 
 import json
 import re
@@ -284,12 +283,9 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals, places=None):
         tokens [Token]: The text to parse.
         short_scale boolean:
         ordinals boolean:
-        places [int] or None: Number of decimal places to return
+        places [int] or None: Number of decimal places to return.
+                              0 truncates the decimal part
                               None performs no rounding
-                              Positive int rounds to so many places
-                              0 value rounds up to nearest int
-                              -1 value rounds down to nearest int
-                              other values throw error
 
     Returns:
         (float, [Token])
@@ -336,22 +332,19 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals, places=None):
             if "." not in str(numbers2[0].text):
                 return_value = float('0.' + "".join([str(
                     decimal.value) for decimal in numbers2]))
-                return_value = number.value + return_value
-                if places is not None:
-                    if places == 0:
-                        return_value = ceil(return_value)
-                    elif places == -1:
-                        return_value = floor(return_value)
-                    if places < 1:
-                        return_value = int(return_value)
                 return_tokens = number.tokens + partitions[1]
                 for n in numbers2:
                     return_tokens += n.tokens
-                if not places:
-                    return return_value, return_tokens
+                if places is not None:
+                    if places > 0:
+                        return_value = number.value + return_value
+                    else:
+                        return_value = number.value
+                else:
+                    return return_value + number.value, return_tokens
 
                 return (round(return_value, places) if places > 0
-                        else return_value), return_tokens
+                        else str(return_value).split('.')[0]), return_tokens
     return None, None
 
 
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -14,16 +14,17 @@
 # limitations under the License.
 #
 from datetime import datetime, timedelta
-
 from dateutil.relativedelta import relativedelta
+from math import ceil, floor
+
+import json
+import re
 
 from lingua_franca.lang.parse_common import is_numeric, look_for_fractions, \
     invert_dict, ReplaceableNumber, partition_list, tokenize, Token, Normalizer
 from lingua_franca.lang.common_data_en import _ARTICLES_EN, _NUM_STRING_EN, \
     _LONG_ORDINAL_EN, _LONG_SCALE_EN, _SHORT_SCALE_EN, _SHORT_ORDINAL_EN
 
-import re
-import json
 from lingua_franca import resolve_resource_file
 from lingua_franca.time import now_local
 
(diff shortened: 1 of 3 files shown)


ChanceNCounter:
found bug: multiple instances of "point" or "dot" in the input string confuse the parser.

lingua_franca/lang/parse_en.py
```diff
@@ -281,24 +289,48 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals):
...
+                    stop = True
+                else:
+                    idx += 1
+            numbers2 = numbers2[:idx]
+
             number = numbers1[-1]
-            decimal = numbers2[0]
+            # decimal = numbers2[0]
 
             # TODO handle number dot number number number
```
krisgesling: Does this PR satisfy this TODO?

lingua_franca/lang/parse_en.py
```diff
@@ -281,24 +289,48 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals):
...
+            if "." not in str(numbers2[0].text):
+                return_value = float('0.' + "".join([str(
+                    decimal.value) for decimal in numbers2]))
+                return_value = number.value + return_value
+                if return_value == int(return_value):
+                    return_value = int(return_value)
+
+                # out_part2 = partitions[2]
+                # for n in numbers2:
+                #     out_part2[n.index] = n.value
```
krisgesling: I'm guessing this is not needed and can be deleted

lingua_franca/lang/parse_en.py
```diff
@@ -281,24 +289,48 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals):
...
             # TODO handle number dot number number number
-            if "." not in str(decimal.text):
-                return number.value + float('0.' + str(decimal.value)), \
-                    number.tokens + partitions[1] + decimal.tokens
+            if "." not in str(numbers2[0].text):
+                return_value = float('0.' + "".join([str(
+                    decimal.value) for decimal in numbers2]))
+                return_value = number.value + return_value
+                if return_value == int(return_value):
+                    return_value = int(return_value)
```
krisgesling: If I was calling an `extract_decimal` function I'd expect a float is returned. What's the benefit of returning an integer instead?

lingua_franca/lang/parse_en.py
```diff
@@ -281,24 +289,48 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals):
...
+                                              ordinals, fractional_numbers=False,
+                                              places=places)
             if not numbers1 or not numbers2:
                 return None, None
 
+            idx = 1
+            stop = False
+            while idx < len(numbers2) and not stop:
+                if numbers2[idx].tokens[0].index != numbers2[idx-1].tokens[0].index + 1 or \
+                        numbers2[idx].value is None:
```
krisgesling: Wonder if a little comment to explain what's going on here would be worthwhile?
Or assigning the elements to descriptive variable names

lingua_franca/lang/parse_en.py
```diff
@@ -77,7 +77,7 @@ def generate_plurals_en(originals):
 _STRING_LONG_ORDINAL_EN = invert_dict(_LONG_ORDINAL_EN)
 
 
-def _convert_words_to_numbers_en(text, short_scale=True, ordinals=False):
+def _convert_words_to_numbers_en(text, short_scale=True, ordinals=False, places=None):
```
krisgesling: I can definitely understand shortening this to `places` in the function that is specifically handling decimal places, however here it's not 100% clear what places would refer to, so may be worth using the full `decimal_places` in these function definitions. 

Be good to update these docstrings while we're at it too.

krisgesling (commented):
Hey Chance, loving all your work on this stuff.

A few questions and suggestions here, but push back if you disagree :)

lingua_franca/lang/parse_en.py
```diff
@@ -281,24 +289,48 @@ def _extract_decimal_with_text_en(tokens, short_scale, ordinals):
...
+
             number = numbers1[-1]
-            decimal = numbers2[0]
+            # decimal = numbers2[0]
 
             # TODO handle number dot number number number
-            if "." not in str(decimal.text):
-                return number.value + float('0.' + str(decimal.value)), \
-                    number.tokens + partitions[1] + decimal.tokens
+            if "." not in str(numbers2[0].text):
```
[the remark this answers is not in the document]
ChanceNCounter: By this point,` _DECIMAL_MARKER` has done its job. This bit is easier to look at in a debugger than it is to explain, but `numbers2` is a list of `ReplaceableNumber` objects corresponding to the digits of our decimal part. It's going to be joined, then appended to a decimal point, then cast to a float and summed with the whole number part of our result.

The `if` statement here goes back to Core, and it makes sure the first element of `numbers2` doesn't already have a decimal point. It's hypothetically possible, but I can't figure out how, and removing the `if` clause doesn't break any test cases. Still, I left it there, just removed a superfluous variable is all.

I'll see if I can work backwards through `git blame` and ask whoever wrote the if statement. I was assuming it had something to do with the way the tokenizer handles certain input. If it's just a relic, I'm all for removing line 318 entirely.

lingua_franca/lang/parse_en.py
ChanceNCounter: Should be `and not places`, good catch! That is, this is fundamentally a helper invoked by `extract_number()`. Unless the user has asked for decimal places, I'd figured no rounding means a whole number is just that. On the other hand, you're right, that doesn't work either: "one" is a whole number. "one point zero" *equals* a whole number.

What if `decimal_places=None` does no rounding, `decimal_places=0` always returns an int, and `decimal_places > 0` does what it does?

krisgesling: :+1: Yeah that sounds intuitive to me

lingua_franca/lang/parse_en.py
```diff
@@ -1487,11 +1542,12 @@ def extract_numbers_en(text, short_scale=True, ordinals=False):
             is now common in most English speaking countries.
             See https://en.wikipedia.org/wiki/Names_of_large_numbers
         ordinals (bool): consider ordinal numbers, e.g. third=3 instead of 1/3
+        decimal_places (int or False): rounds to # decimal places. uses builtin round()
```
krisgesling, line 1545: One more `int or False`

lingua_franca/parse.py
```diff
@@ -90,12 +94,19 @@ def extract_numbers(text, short_scale=True, ordinals=False, lang=None):
             See https://en.wikipedia.org/wiki/Names_of_large_numbers
         ordinals (bool): consider ordinal numbers, e.g. third=3 instead of 1/3
         lang (str): the BCP-47 code for the language to use, None uses default
+        decimal_places (int or None): Positive value will round to X places.
+                                      Val of 0 will round up to nearest int,
+                                        equivalent to `math.ceil(result)`
+                                      Val of -1 will round down to nearest int,
+                                        equivalent to `math.floor(result)`
+                                      Val of None will perform no rounding,
+                                      potentially returning a very long string.
```
krisgesling, line 103: Was this a previous idea for implementation? Assuming this will be the new behaviour you described where 0 is a regular rounding and there is no handling of a `-1` input?

ChanceNCounter: At this point, I can't decide which way is better. There are three things at play:

* If these functions don't round, their output could be a string of stupendous length
* You don't always want to round the same way
* Sometimes you feel like an int, sometimes you don't

ChanceNCounter:
The status box on this PR seems to be broken. If I click on Travis, it shows all the builds completed successfully.

forslund:
It happens sometimes with the checks, if there's a communication error on the webhook the status update can be lost.

lingua_franca/lang/parse_en.py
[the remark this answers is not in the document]
JarbasAl: if it doesnt break test cases, i say let's remove it

PR #94, opened 2020-03-19
extract single months

### LINKED ISSUE #20: extract_datetime fails when only month supplied
If an utterance like "when is september" is parsed using extract_datetime, it results in:
> ValueError: time data 'september' does not match format '%B %d %Y'

There doesn't seem to be an easy answer to what should be returned as this would be dependent on other contextual factors eg 
- On 29 August 2019 - "when is september" might be 1 September 2019.
- On 29 December 2019 - "when is september" might be 1 September 2020.
- On 29 December 2019 - "when was september" might be 30 September 2019.

So thought it was worth opening up the conversation on how this situation should be handled?

JarbasAl:
closes #20

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -236,10 +236,12 @@ def _extract_fraction_with_text_en(tokens, short_scale, ordinals):
         if len(partitions) == 3:
             numbers1 = \
                 _extract_numbers_with_text_en(partitions[0], short_scale,
-                                              ordinals, fractional_numbers=False)
+                                              ordinals,
+                                              fractional_numbers=False)
             numbers2 = \
                 _extract_numbers_with_text_en(partitions[2], short_scale,
-                                              ordinals, fractional_numbers=True)
+                                              ordinals,
+                                              fractional_numbers=True)
 
             if not numbers1 or not numbers2:
                 return None, None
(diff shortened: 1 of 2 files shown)


lingua_franca/lang/parse_en.py
```diff
@@ -635,6 +639,38 @@ def extract_duration_en(text):
     return (duration, text)
 
 
+def month_to_int_en(month):
+    if isinstance(month, int) or isinstance(month, float):
+        return int(month)
+    if isinstance(month, str):
```
ChanceNCounter: I think this should become a lang-specific dictionary, and the function should move upstairs, so that each parser can just do something like, `month_to_int(word, lang_code)`

lingua_franca/lang/parse_en.py
```diff
@@ -882,12 +918,14 @@ def date_found():
                 start -= 1
                 # parse 15 of July, June 20th, Feb 18, 19 of February
         elif word in months or word in monthsShort and not fromFlag:
-            try:
-                m = months.index(word)
-            except ValueError:
-                m = monthsShort.index(word)
+            m = month_to_int_en(word)
+
+            if wordPrev in ["next", "this", "last", "past", "previous"]:
```
ChanceNCounter: Looks like "this <week/month/year>" still returns nothing. Idk if that's in scope.

lingua_franca/lang/parse_en.py
```diff
@@ -1383,29 +1421,53 @@ def date_found():
...
+                temp = datetime.strptime(datestr, "%B %d %Y")
+            except:
+                temp = None
+        if temp is None:
+            # check for month
+            for m in months + monthsShort:
+                if m in datestr:
+                    int_month = month_to_int_en(m)
+                    year = extractedDate.year
+                    if int_month < extractedDate.month:
```
ChanceNCounter: * "March 15th" returns Mar 15 of next year (OK)
* "Next March 15th" returns Mar 1 of THIS year, (which, in addition to being the wrong date, already happened)
* "This March 15th" also returns Mar 1 of this year
* "Last March 15th" returns March 1 of last year

Also, how would one (intentionally) ask for the March 15th that just happened?

PR #95, opened 2020-03-20
refator/extract_duration

JarbasAl:
refactor of extract_duration:
- support durations in months/years/decades/centuries/millenniums
- add optional replace_token param, allows better parsing of remainder_string
	- used in #96 
- add a DateResolution parameter
	- extracted duration can be in [datetime.timedelta](https://docs.python.org/3/library/datetime.html)
		- months/years use the average length of days in a month/year
	- extracted duration can be in [dateutil.relativedelta](https://dateutil.readthedocs.io/en/stable/relativedelta.html)
		- months/years are relative
	- extracted duration can be in microseconds/milliseconds/seconds
	- extracted duration can be in minutes/hours/days/weeks
	- extracted duration can be in months/years/decades/centuries/millenniums
		- months/years use the average length of days in a month/year


lingua_franca/lang/parse_en.py
```diff
@@ -629,10 +634,35 @@ def extract_duration_en(text):
...
         text = re.sub(unit_pattern, '', text)
 
-    text = text.strip()
-    duration = timedelta(**time_units) if any(time_units.values()) else None
+    extended_units = {
+        'months': 0,
+        'years': 0,
+        'decades': 0,
+        'centurys': 0,
+        'millenniums': 0
```
ChanceNCounter: Incidental, but it's "millennia" from Latin.

JarbasAl: both are correct and both should be supported

ChanceNCounter:
Migrating discussion from #95, re: number of days in a month:

I was indeed referring to `datetime.date`, which supports years, months, and days, and can be made to play with datetimes themselves, as well as timedeltas.

I expect that, most of the time, a person who asks for "in a month" from today will want April 23, not April 22. If they ask for 30 days, otoh...

lingua_franca/lang/parse_en.py
```diff
@@ -342,7 +346,8 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
         next_word = tokens[idx + 1].word if idx + 1 < len(tokens) else ""
 
         if is_numeric(word[:-2]) and \
-                (word.endswith("st") or word.endswith("nd") or word.endswith("rd") or word.endswith("th")):
+                (word.endswith("st") or word.endswith("nd") or
```
ChanceNCounter: As long as this bit's turned green, I'd like to evangelize for `any()` and `all()` wherever they're logical, because much of our code lives in loops, over input of arbitrary length.

```python
has_ordinal_suffix = (word.endswith(suffix) for suffix in ["st", "nd", "rd", "th"])
any(has_ordinal_suffix)
```

just as an "as long as you're in there" thing. It looks tiny every time we do it, but it'll add up. Doesn't work everywhere, but it works in many places.

ChanceNCounter:
> I expect that, most of the time, a person who asks for "in a month" from today will want April 23, not April 22. If they ask for 30 days, otoh...

I remember now. It's been a long time, had to poke around. It's this I was thinking of:

```python
>>> foo = datetime.datetime.now() + dateutil.relativedelta.relativedelta(months=+1)
>>> foo
datetime.datetime(2020, 4, 23, 21, 0, 31, 607268)
>>> foo - datetime.datetime.now()
datetime.timedelta(days=30, seconds=86395, microseconds=214964)

**Comment by JarbasAl** [comment]
extract_duration is not used exclusively in datetime contexts, it means a certain length of time and SI units are important. If i am for some reason talking about the total length of my music collection and that is >1 month this behaviour would be strange... 

You do have a valid point about relativedeltas making more sense in the date context, i will instead allow a flag to decide if you want a relativedelta or a timedelta returned.  

Changing the return type would be a breaking change, using a flag also makes it easier to maintain compatibility. For backwards compat this flag needs to be off by default, but we should look into making on be default in the future, a relativedelta should behave the same, but type checks will break existing code

On a different note, a timedelta is not necessarily the best type for a duration in a non date context, this flag could also support seconds, so maybe i can just add a Resolution enum?

File: lingua_franca/lang/parse_en.py (hunk 2 above)
Comment (JarbasAl) [reply lingua_franca/lang/parse_en.py]: i think that in general those changes should be part of a different PR (code cleanup PR?), these green bits are just pep8 fixes pycharm does automatically.... if we are touching actual logic i'd rather have it reviewed on it's own rather than lost in an unrelated feature. We want to make it easy to track and revert changes that cause bugs, Unmerge PR is easier than looking for specific commits. so while this is a tiny thing and relatively harmless it's just a better practice in general.

**Comment by ChanceNCounter** [comment]
I guess what concerns me is the situations where the parsers would layer. Take Mycroft's timer skill, for example, or a reminder skill. If I ask it to "remind me in a month," is that going to extract today's datetime, find a relative delta for "one month later," and schedule the timer on what (as of today) would be April 24?

Or, will it extract the duration "one month," call that 30 days, and set a reminder to fire on April 23?

File: lingua_franca/lang/parse_common.py [hunk 3]
```diff
@@ -385,3 +386,20 @@ def replace_right(source, target, replacement, replacements=None):
             extract = extract_handler(to_parse, short_scale, ordinals)
     numbers.reverse()
     return numbers
+
+
+class DurationResolution(Enum):
```
ChanceNCounter: Could this move "up" to common, as TimeResolution, maybe add the "delta" members to an override in here?

We're going to want the same enum in the date and duration *formatters* down the road.

```
class TimeResolution(Enum):
    MICROSECONDS = 0
    SECONDS = 1
    MINUTES = 2
    ...

**Comment by JarbasAl** [comment]
> I guess what concerns me is the situations where the parsers would layer. Take Mycroft's timer skill, for example, or a reminder skill. If I ask it to "remind me in a month," is that going to extract today's datetime, find a relative delta for "one month later," and schedule the timer on what (as of today) would be April 24?
> 
> Or, will it extract the duration "one month," call that 30 days, and set a reminder to fire on April 23?

To be totally correct, a timer skill should use the timedelta (remind me IN a month), a calendar skill should use a relativedelta (remind me NEXT month)

Either way it would be the responsibility of skill to set the resolution flag as appropriate

Layering should be without issue, most operations in date/datetime objects will just work since the timedelta and relativedelta objects behave more or less the same

**Comment by ChanceNCounter** [comment]
Right, but then those function calls won't produce the natural result. "In one month" will almost always intend a relativedelta from datetime.now(), rather than a fixed timedelta, but it will often be fed to intents that would otherwise call `extract_duration()`: utterances containing things like, "for three hours", "to six weeks," "in ninety minutes."

I certainly agree that changing the return type is a bad move, but why not hinge it on the input type/s? extract_datetime() and perform substitution, if there's now a datetime, return a timedelta *via* a relativedelta. If not, fall back on fixed values.

If extract_duration doesn't account for the calendar, I suspect it'll end up clobbering input data on the way through skills. It's the skill's responsibility to use the right parser, sure, but you've gotta be able to pass tokens around and expect consistent results.

**Comment by JarbasAl** [comment]
> Right, but then those function calls won't produce the natural result. "In one month" will almost always intend a relativedelta from datetime.now(), rather than a fixed timedelta, but it will often be fed to intents that would otherwise call `extract_duration()`: utterances containing things like, "for three hours", "to six weeks," "in ninety minutes."
> 
> I certainly agree that changing the return type is a bad move, but why not hinge it on the input type/s? extract_datetime() and perform substitution, if there's now a datetime, return a timedelta _via_ a relativedelta. If not, fall back on fixed values.
> 
> If extract_duration doesn't account for the calendar, I suspect it'll end up clobbering input data on the way through skills. It's the skill's responsibility to use the right parser, sure, but you've gotta be able to pass tokens around and expect consistent results.

since a relative delta is equivalent to a timedelta if < months, what you are describing is basically setting DateResolution.RELATIVEDELTA as the default value. 

Setting this flag behaves the way you are describing. I'm only not making it a default for backwards compatibility but i don't think there is any case in the wild doing type checks, so setting this as default should be perfectly safe

I still think that a consistent result means that DateResolution.TIMEDELTA should behave like it does / that is the natural result for the meaning of the word "duration". a natural result in this case conflicts with a correct result, durations can not depend on reference.

I'd say that since the majority of use cases for this should be with dates, setting DateResolution.RELATIVEDELTA  as a default is a good compromise. But i want to keep the SI units approach in there

**Comment by ChanceNCounter** [comment]
Makes sense to me!

**Comment by JarbasAl** [comment]
there is one big issue, RELATIVEDELTA can not be the default

what does 1.2 months ago mean? apart from a `TypeError` 0.2 months would mean different things depending on reference_date, TIMEDELTA is the only sane choice for a default, that is the meaning of duration...

to handle this in the case of DurationResolution.RELATIVEDELTA i can:
- throw an exception (let users handle it)
- log a warning and fallback to DurationResolution.TIMEDELTA (change of paradigm)
- approximate, decimal part can be converted to days 

demonstration of type error
```python
from dateutil.relativedelta import relativedelta
from datetime import datetime, timedelta


now = datetime.now()


d = relativedelta(seconds=1.3)
seconds_ago = now - d

d = relativedelta(minutes=1.3)
minutes_ago = now - d

d = relativedelta(hours=1.3)
hours_ago = now - d

d = relativedelta(days=1.1)
day_ago = now - d

d = relativedelta(weeks=1.3)
week_ago = now - d

d = relativedelta(days=120)
day_ago = now - d

#  exceptions bellow
d = relativedelta(microsecond=1.0)
microsecond_ago = now - d


d = relativedelta(months=1.0)
month_ago = now - d


d = relativedelta(year=1.0)
year_ago = now - d

```

ChanceNCounter:
Maybe go with TIMEDELTA for the default, and then add another top-level function `extract_calendar_duration()` for the use cases we discussed above, which will just call `extract_duration(resolution=RELATIVEDELTA)`.

Standard docstring "equivalent to..." and voila.

ChanceNCounter:
The need to rebase aside, was this ready to go?

lingua_franca/lang/parse_en.py
```diff
@@ -580,7 +586,8 @@ def extractnumber_en(text, short_scale=True, ordinals=False):
                                         short_scale, ordinals).value
 
 
-def extract_duration_en(text):
+def extract_duration_en(text, resolution=DurationResolution.TIMEDELTA,
```
ChristopherRogers1991, line 589: This function now conflates two distinct ideas - durations and offsets. I would propose splitting this into two distinct functions - `extract_duration_en`, which always returns a `timedelta` (see my comment below about unit conversions), and `exract_offset_en` which always returns a `relativedelta`. The common logic can be extracted out into small, helper functions.

Note that this is basically the distinction that is made in `parse.py`, so being consistent here and using two functions for those two different concepts seems like it would make a lot of sense.

lingua_franca/lang/parse_en.py
```diff
@@ -609,30 +618,229 @@ def extract_duration_en(text):
...
+        if resolution == DurationResolution.TOTAL_MICROSECONDS:
+            duration = microseconds
+        elif resolution == DurationResolution.TOTAL_MILLISECONDS:
+            duration = microseconds / 1000
+        elif resolution == DurationResolution.TOTAL_SECONDS:
+            duration = microseconds / (1000 * 1000)
+        elif resolution == DurationResolution.TOTAL_MINUTES:
+            duration = microseconds / (1000 * 1000 * 60)
+        elif resolution == DurationResolution.TOTAL_HOURS:
+            duration = microseconds / (1000 * 1000 * 60 * 60)
```
ChristopherRogers1991, line 819: Using the `unit_to_multiplier` map suggested above would save defining these twice. You could just use `duration = microseconds / unit_to_multiplier[unit]` here.

lingua_franca/lang/parse_common.py
```diff
@@ -385,3 +386,23 @@ def replace_right(source, target, replacement, replacements=None):
...
```
ChristopherRogers1991, line 391: I don't think this enum should exist (or should at least be reduced in scope/split up). As noted in my comment below, the changes in the PR conflate the concepts of durations and offsets. I think it would be better to use two separate functions rather than using this enum to change the behavior _and_ return type of a single function. Separating the functions should result in cleaner logic, and consistent return types per function.

lingua_franca/lang/parse_common.py
```diff
@@ -385,3 +386,23 @@ def replace_right(source, target, replacement, replacements=None):
...
+    TIMEDELTA = 0
+    RELATIVEDELTA = 1
+    RELATIVEDELTA_STRICT = 1
```
ChristopherRogers1991, line 394: You have two `1`s here. If these are actually equivalent, I would suggest just using one, rather than having two equal values.

Not sure what the target python version is here, but if it's 3.6 or better, I would recommend using [auto](https://docs.python.org/3.8/library/enum.html#enum.auto).

lingua_franca/parse.py
```diff
@@ -158,7 +159,33 @@ def extract_number(text, short_scale=True, ordinals=False, lang=None):
...
+     Args:
+        text (str): string containing a duration
+        lang (str): the BCP-47 code for the language to use, None uses default
+        replace_token (str): string to replace consumed words with
+
+    Returns:
+        (relativedelta, str):
+                    A tuple containing the duration and the remaining text
+                    not consumed in the parsing. The first value will
+                    be None if no duration is found. The text returned
```
ChristopherRogers1991, line 177: > The text returned will have whitespace stripped from the ends.

This isn't always true - `strip()` is only called if the `replace_token` is set.

lingua_franca/parse.py
```diff
@@ -158,7 +159,33 @@ def extract_number(text, short_scale=True, ordinals=False, lang=None):
...
+    """
+    return extract_duration(text, lang,
+                            DurationResolution.RELATIVEDELTA_FALLBACK,
+                            replace_token)
+
+
+def extract_duration(text, lang=None,
+                     resolution=DurationResolution.TIMEDELTA,
+                     replace_token=""):
     """ Convert an english phrase into a number of seconds
```
ChristopherRogers1991, line 189: These docs are incorrect - this does not return a number of seconds (the new version has multiple types, and the old version returns a `timedelta`).

lingua_franca/lang/parse_en.py
```diff
@@ -609,30 +618,229 @@ def extract_duration_en(text):
...
+                value = relative_units[unit]
+                _leftover, _ = math.modf(value)
+                if _leftover != 0:
+                    raise ValueError("relativedelta requires {unit} to be an "
+                                     "integer".format(unit=unit))
+                relative_units[unit] = int(value)
+        duration = relativedelta(**relative_units) if \
+            any(relative_units.values()) else None
+    else:
+        microseconds = 0
```
ChristopherRogers1991, line 769: Why not just use the logic from above that generates the `timedelta`, and then use `timedelta.total_seconds() * 1000000` to get microseconds?

lingua_franca/parse.py
```diff
@@ -158,7 +159,33 @@ def extract_number(text, short_scale=True, ordinals=False, lang=None):
     return text
 
 
-def extract_duration(text, lang=None):
+def extract_calendar_duration(text, lang=None, replace_token=""):
```
ChristopherRogers1991, line 162: Would recommend calling this `extract_offset`, or `extract_relativedelta`.

lingua_franca/lang/parse_en.py
```diff
@@ -609,30 +618,229 @@ def extract_duration_en(text):
...
+                if si_units["days"] is None:
+                    si_units["days"] = 0
+                si_units["days"] += 100 * DAYS_IN_1_YEAR * value
+            elif unit == "millenniums":
+                if si_units["days"] is None:
+                    si_units["days"] = 0
+                si_units["days"] += 1000 * DAYS_IN_1_YEAR * value
+            else:
+                si_units[unit] = value
+        duration = timedelta(**si_units) if any(si_units.values()) else None
```
ChristopherRogers1991, line 684: `any(si_units.values())` is the same whether the values in `si_units` are initialized to `None` or `0` (it will be `False`).

If you initialize to `0`, you don't have to do all the `None` checks in the nested `if`s, which cleans up the code a little.

In either form though, `0` is not a valid duration (e.g. '0 seconds' would return `None`), which seems incorrect - I think it should return `0`. I'd recommend rethinking this logic, to account for that case. The best move is probably to calculate everything in terms of one unit, e.g. seconds, and then just use that to create the `timedelta`, rather than using the dict with distinct values for each unit.

ChristopherRogers1991 (commented):
I wrote the [original extract_duration](https://github.com/MycroftAI/mycroft-core/pull/1977/files), so when I saw it was being refactored, I figured I'd take a peak. Left some comments about some things I think could be cleaned up. It's kind of late for me, so... hopefully they all make sense. Feel free to reach out if they don't.

lingua_franca/lang/parse_en.py
```diff
@@ -609,30 +618,229 @@ def extract_duration_en(text):
...
+                    relative_units["years"] = 0
+                relative_units["years"] += value * 100
+            elif unit == "millenniums":
+                if relative_units["years"] is None:
+                    relative_units["years"] = 0
+                relative_units["years"] += value * 1000
+            else:
+                relative_units[unit] = value
+
+        # microsecond, month, year must be ints
```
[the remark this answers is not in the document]
JarbasAl: what does 1.5 months mean?  a month is anywhere from 28 to 31 days. Same goes for year depending on on february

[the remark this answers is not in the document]
JarbasAl: this is the main distinction between delta and relativedelta

lingua_franca/lang/parse_en.py
```diff
@@ -609,30 +618,229 @@ def extract_duration_en(text):
...
-        time_units[unit] = value
-        text = re.sub(unit_pattern, '', text)
-
-    text = text.strip()
-    duration = timedelta(**time_units) if any(time_units.values()) else None
-
-    return (duration, text)
+    text = text.replace("centuries", "century").replace("millenia",
+                                                        "millennium")
+    text = text.replace("a day", "1 day").replace("a year", "1 year") \
```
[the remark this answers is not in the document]
JarbasAl: not sure, its been a while since i touched this, i think _convert_words_to_numbers_en missed these

lingua_franca/parse.py
ChanceNCounter: I'm all for something less clunky than `extract_calendar_duration`, but I'd like it if all the functions in LF had such painfully descriptive names, somebody can figure them out who has never written code before.

JarbasAl:
@ChristopherRogers1991  great feedback, will implement most of those changes and rebase this PR

PR #96, opened 2020-03-22
Refactor/date extract

### LINKED ISSUE #20: extract_datetime fails when only month supplied

### LINKED ISSUE #49: Recognize holidays by name
This is a bit of a meta-issue, since it's quite complicated. In many cases of natural speech we refer to days by a name such as Christmas or Tuesday or black friday. The library already has some names like the days of the week, but Google has a much larger database of entities that they can recognize, including holidays and named date ranges.

* when is columbus day
* how many days until mother's day?
* play bing crosby music at 10am on christmas day
* when is black friday?
* how many days until the start of spring?
* turn off sprinklers next week.
* turn off sprinklers until Monday

Holidays are often culture or country-specific rather than language-specific (e.g., Thanksgiving is a different day in Canada than in the US, and mother's day is different in the U.K.). They can also be complicated to calculate (easter sunday is a good example). There is also a [huge tail of dates](https://www.timeanddate.com/holidays/us/) that have names associated with them. If you try to cover everything then you will descend into a maze of bugs (e.g., Google gets it wrong when you ask how many days until winter, and I once filed a bug for Columbus day). Rather than getting too concerned with the rare or hard cases, it would be useful to start a database of easily calculated holidays that can be recognized and converted to a date, but allow for some extensibility and customizability based on country or personalization (e.g., "Dad's birthday", or "spring break" from your own calendar).

The scope of this issue needs to be defined precisely in terms of what is output. I would restrict it to only dates or date ranges that can be unambiguously determined in the context. This is highly non-trivial, since even things like "next week" or "next friday" are ambiguous. The existing library seems to have made arbitrary decisions for these ambiguities, preferring the most common usage.

### LINKED ISSUE #59: extract_datetime_en() fails silently in several cases, returns wrong datetime
https://github.com/MycroftAI/mycroft-core/blob/a3e35d736b0426f33b7fc1d1883be38e4b650b0b/mycroft/util/lang/parse_en.py#L667

Several cases not yet implemented and consequently parsing incorrectly:

- [ ]  **"Yesterday**: this one's probably somewhat urgent - may constitute a bug rather than unimplemented functionality
- [ ]  "of <this/next/last> <year/month/week>"
- [ ]  "# days <before X/after X/ago>" currently parses to "# days after today"

### LINKED ISSUE #72: extract_datetime_en() clobbers any additional datetimes in string
```python
>>> parse.extract_datetime("september 30 2019 october 24th 2018")
[datetime.datetime(2018, 10, 24, 0, 0, tzinfo=tzlocal()), '']
```
Desirable output would be:

```python
[datetime.datetime(2018, 10, 24, 0, 0, tzinfo=tzlocal()), 'september 30 2019']
```

It might be worth adding an extract_datetimes() function while fixing this, or the patcher could regard that as out of scope. They're the same problem space, this bug prevents that potential feature.

### LINKED ISSUE #74: Possible extract_datetime rewrite
My TODO list, and now this issue tracker, are becoming cluttered with bugs and missing functionality in the datetime parsers, and I'm beginning to wonder if we'd be better-served rewriting them to recurse over anchors.

That is, right now, it tries to parse input strings a word at a time. This catches most cases in English, but I've just found a not-so-edge case: "day after tomorrow" is great, but "three days after tomorrow" returns... tomorrow.

"December 3rd at 3pm" returns correctly. So does "3pm on December 3rd." However, "December 3rd at three pm" returns dec 3 with a remainder of "three pm." "December third" throws that formatting error.

The parser is turning into whack-a-case.

---

I'd like to open a discussion about recursively pulling keywords and offsets from the input string. For example, if the string contains "tomorrow," the rest of our parsing is relative to tomorrow's date. We divide the string at that word, and then check the strings before and after for more relative information.

```
"three o'clock tomorrow afternoon and again at five the next morning" --->
["three o'clock", tomorrow_date, "and again at five the next morning"]

"three o'clock" + tomorrow_date --->
[tomorrow_datetime(3am), " afternoon and again at five the next morning"]

[tomorrow_datetime(3am), "afternoon", "and again at five the next morning"]
tomorrow_datetime(3am) + "afternoon" --->
[tomorrow_datetime(3pm), "and again at five the next morning"]
```

---

In the above pseudocode, the hardest bit would be recognizing that the remainder, though it contains datetime-related information, is *another* datetime, and should not be handled by extract_datetime(), but rather by whatever is calling the function. In our case, that would likely be an extract_datetimes() function.

If we can nail down logic like that, however, that continually digests the input, rather than going word by word, it should be easier to code around edge cases, with an end goal being

`[offset: ["three days after", 3pm], entity: [scheduled_event: "the superbowl"]]` ---> `datetime(three days after superbowl sunday at 3pm)`

JarbasAl:
Complete rewrite of extract_datetime, 

this is a work in progress, DO NOT MERGE. 

Opening PR to generate discussion / collaboration

- split extract_datetime into
	- extract_date
	- extract_time

- adds location awareness
   - support for seasons (summer/winter/spring/fall)
   - support for location based holiday parsing (+ global names dates)
- adds DateResolution for granularity control of ambiguous dates
- adds a bunch of time utils for parsing date_ranges
- "now" always refers to current / execution time, else reference_date is used (today == reference_date, now == execution_date)

NOTE: depends on/includes #95 
- closes #20 
- closes #49 
- closes #59 
- **WIP** closes #72 
- partial solution for #74


lingua_franca/lang/parse_en.py
```diff
@@ -671,15 +672,15 @@ def extract_duration_en(text, replace_with=""):
         text = re.sub(unit_pattern, replace_with, text)
 
         if unit == "months":
-            time_units["days"] += 30 * value
+            time_units["days"] += DAYS_IN_1_MONTH * value
```
ChanceNCounter: This isn't a consistent figure. I think the best route is to rely on the datetime module, increment the `month` attribute, and distinguish between '30 days' and 'one month'.

JarbasAl: the datetime module provides timedelta objects, but the highest resolution they provide is in days. This makes sense since we can define anything from days and bellow in terms of SI units (seconds). 

Since this represents a time interval (a duration) it must mean the same in whatever reference date. if a month has a different duration depending on reference date it can not be used for measurement of an interval

Silly example, imagine watching a video that randomly lasted 28 or 31 days....

JarbasAl: also note, this belong in pr#95 which is required for this PR, please move discussion there

ChanceNCounter:
May close #59

JarbasAl:
adding support for ```{nth} {unit} since {start_date}```

**WIP**:
 -  Coordinated Universal Time, adopted 1960 and since 1972 including a system of observation-based leap seconds.
 -  ISO 8601, standard based on the Gregorian calendar, Coordinated Universal Time and ISO week date, a leap week calendar system used with the Gregorian calendar
 -  Fiscal year varies with different countries. (location aware)
 -  Unix time, number of seconds elapsed since 1 January 1970, 00:00:00 (UTC).
 -  Julian day, number of days elapsed since 1 January 4713 BC, 12:00:00 (UTC).
 -  Heliocentric Julian Date, Julian day corrected for differences in the Earth's position with respect to the Sun.
 -  Barycentric Julian Date, Julian day corrected for differences in the Earth's position with respect to the barycentre of the Solar System.
 -  Lilian date, number of days elapsed since the beginning of the Gregorian Calendar on 15 October 1582.
 -  Rata Die, number of days elapsed since 1 January 1 AD 1 in the proleptic Gregorian calendar.

JarbasAl:
now that the big refactor is in i am getting back to this one

PR #97, opened 2020-03-25
Refactor/languages cleanup

JarbasAl:
- move all language constants to common_XX
- UPPERCASE all module constants
- unCamelCase function names
- unify function signatures across all languages
- add templates that raise `NotImplementedError` for all missing languages
	- only for languages that have at least 1 method implemented
- add lang agnostic methods where missing in parse and format, example `is_fractional` and `nice_part_of_day`
- add `_langcode` sufix to all lang specific methods
- add `_underscore` prefix to all not officially exported methods/CONSTANTS
- random pep8 tweaks


lingua_franca/format.py
```diff
@@ -22,22 +22,11 @@
...
-from lingua_franca.lang.format_fr import nice_number_fr
-from lingua_franca.lang.format_fr import nice_time_fr
-from lingua_franca.lang.format_fr import pronounce_number_fr
-from lingua_franca.lang.format_nl import nice_time_nl
-from lingua_franca.lang.format_nl import pronounce_number_nl
-from lingua_franca.lang.format_nl import nice_number_nl
-from lingua_franca.lang.format_da import nice_number_da
-from lingua_franca.lang.format_da import nice_time_da
-from lingua_franca.lang.format_da import pronounce_number_da
+from lingua_franca.lang.format_es import *
```
ChanceNCounter: If this bit is getting refactored, let's reduce the number of changes we'll need to make when we add new languages. Instead of importing these functions wholesale and calling them ad-hoc, let's give the parser calls something like:

```python
def nice_response(text, lang=None):
    lang_code = get_primary_lang_code(lang)
    if lang_code:
        format_lib = importlib.import_module("lingua_franca.lang.format_" + lang_code)
        _nice_response = getattr(format_lib, "nice_response_" + lang_code)
        return _nice_response(text)
    ...

**Comment by ChanceNCounter** [comment]
I love this. I'll be reviewing it piecemeal over the course of the evening.

File: lingua_franca/format.py (hunk 1 above)
Comment (ChanceNCounter) [reply lingua_franca/format.py]: In fact, that bit could probably become its own helper function, and the parser calls could pass the strings to it.

```python
def find_format_function(name, lang):
    lang_code = get_primary_lang_code(lang)
    if lang_code:
      return getattr(importlib.import_module("lingua_franca.lang.format_" + lang_code),
                  name + "_" + lang_code)

(written in browser)

Comment (forslund) [reply lingua_franca/format.py]: I'm a bit skeptical to this approach, sure it's nice for people who knows the layout of the project but for new devs it'll look really weird and the traceability of the origin of the functions become really unclear.

easier may be to just import the language module

```python
from .lang import format_en

[....]
    format_en.extract_number_en(...)
```

The language suffix on the functions in each module could be dropped to make the parsing cleaner.

ChanceNCounter: Fair enough, though I think comments could solve that. I guess it'd just be replacing clutter with clutter.

The trouble with the current approach is that it's a thing to miss, a place individual parsers have to be "installed" as they're written. `NotImplementedError` seems friendlier in that respect.

ChanceNCounter: @forslund made the excellent point/suggestion (PM) that this would be handled better by caching function signatures. On import, using importlib, discover each language's parser, and store them in dicts: `extract_number_funcs = {lang_code: function}`.

Then, the top-level functions dynamically call the relevant parsers via those dicts.

If we did this, we'd lose the static imports - no longer we'd no longer be importing languages the user isn't using, nor importing whole files at once - but, unlike what I proposed above, we'd only be going over the localized files *once*, when `lingua_franca` is imported, rather than *every time you call a function* (doh!)

Much more sensible.

ChanceNCounter:
If the dynamic import thing is implemented, I think those templates might be better off raising `NotImplementedError` when discovery fails, so that the functions don't show up in autogenerated docs.

lingua_franca/format.py
ChanceNCounter: Working WIP:

```python3
from importlib import import_module

def load_loc_function(function_name, lang_code):
    try:
        module = import_module(".lang.format_" + lang_code,
                                "lingua_franca")
    except ModuleNotFoundError:
        return None

    try:
        function = getattr(module, function_name + "_" + lang_code)
    except AttributeError:
        return None

    return function

Comment (JarbasAl) [reply lingua_franca/format.py]: feel free to commit directly to this branch, i don't think i will touch it anytime soon @ChanceNCounter @forslund

Comment (ChanceNCounter) [reply lingua_franca/format.py]: Yep! I'll push this to my repo first, so we can diff it against *this* briefly, then bring it in here. Further progress, it works! Holy cow, it works! I have to figure out logic for the `None` situations, but:

```python3
>>> from lingua_franca import common_data, format
>>> pronounce_number_funcs = {lang_code: format.load_loc_function("pronounce_number", \
lang_code) for lang_code in common_data._SUPPORTED_LANGUAGES}
>>> pronounce_number_funcs["en"](357.52)
'three hundred and fifty seven point five two'

**Comment by JarbasAl** [comment]
> If the dynamic import thing is implemented, I think those templates might be better off raising `NotImplementedError` when discovery fails, so that the functions don't show up in autogenerated docs.

on the contrary, we want to document exactly what is missing, basically all languages have partial support only, this makes it clear for end users and people reading the docs it is not implemented. Maybe someone tackles one of them once they get a NotImplementedError.

This also makes it easy to do ```pronounce_number_funcs = {lang_code: format.load_functions("pronounce_number", \
lang_code) for lang_code in common_data._SUPPORTED_LANGUAGES}``` since all methods are guaranteed to exist for any partially implemented language

**Comment by ChanceNCounter** [comment]
That's true. On the other hand, the function loader currently catches those errors if the function doesn't exist. It could be modified to pass the errors through, so that anybody calling them would get a direct printout:

```
NotImplementedError: function 'nice_response' is not implemented for language 'en'.

ChanceNCounter:
Also, it would absolve the people writing new modules of the need to write boilerplate for unimplemented parsers, and it would absolve a person writing a new parser of the need to write boilerplate in every language's file.

lingua_franca/format.py
ChanceNCounter: [Here](https://github.com/MycroftAI/lingua-franca/compare/refactor/languages_cleanup...ChanceNCounter:refactor/languages_cleanup) is a working fork of this branch, where the formatters (but not the parsers) have been moved to that dynamic importer, and are cached.

Using this method, new top-level functions need only be registered in `<module>._REGISTERED_FUNCTIONS` (which could lose its `_`, I guess.) The loader takes care of the rest. `format.py` winds up with:

```python3
 _REGISTERED_FUNCTIONS = ["nice_number",
                         "nice_time",
                         "pronounce_number",
                         "nice_response",
                         "nice_ordinal",
                         "nice_part_of_day"]

_LOCALIZED_FUNCTIONS = common_data.populate_localized_function_dict("format")
```

and that's it! Unimplemented parser calls raise an error. The branch currently contains, for the sake of debugging, an unimplemented *language code* in `common_data._SUPPORTED_LANGUAGES`. This is to demonstrate that such a mistake will throw a warning on import, but won't crash the module.

RFC @JarbasAl @forslund and anybody else who has thoughts.

ChanceNCounter: (the whole thing does rely on uniform function naming, but that shouldn't be hard. we were already pulling it off.)

JarbasAl: i have mixed feelings about the passtrough... the 4 things i feel missing:

- not implemented error should be in auto generated docs, we want to document what is missing, not hide it away
- casually browsing a lang/xxx file does not tell the user what is missing
- the exception traceback does not show the user where to implement the missing functionality
- individual templates could print a message in the target language

```python
def do_something_pt(*args, **kwargs):
    raise NotImplementedError("Esta funcionalidade ainda não foi implementada, tu podes ajudar!")
```

ChanceNCounter: The fork's been heavily updated, and addresses most of this. It doesn't yet address autogenerated docs, and others might prefer a different solution where I'm currently raising exceptions.

Currently, some parsers do not work, and *no* parsers have been migrated. All formatters work, and afaik all the ones that have localized equivalents have been migrated.

Current behavior:

```python3
>>> from lingua_franca import format
>>> format.pronounce_number(65.7, "en")
'sixty five point seven'
>>> format.pronounce_number(65.7, "da")
'femogtres komma syv nul'
g, arguments)
```
```python3
>>> format.nice_response("65.7", "da")
'65.7'
>>> format.nice_response("65.7", "en")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/chance/lingua-franca/lingua_franca/format.py", line 543, in nice_response
    locals().items()) or text
  File "/home/chance/lingua-franca/lingua_franca/format.py", line 53, in call_localized_function
    func_name, lang, arguments)
  File "/home/chance/lingua-franca/lingua_franca/__init__.py", line 56, in _localized_function_caller
    raise func
lingua_franca.FunctionNotLocalizedError: The requested function is not implemented in English.
```
```python3
>>> format.nice_response("65.7", "uk")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/chance/lingua-franca/lingua_franca/format.py", line 543, in nice_response
    locals().items()) or text
  File "/home/chance/lingua-franca/lingua_franca/format.py", line 53, in call_localized_function
    func_name, lang, arguments)
  File "/home/chance/lingua-franca/lingua_franca/__init__.py", line 49, in _localized_function_caller
    raise_unsupported_language(lang_code)
  File "/home/chance/lingua-franca/lingua_franca/__init__.py", line 27, in raise_unsupported_language
    .format(language=language, supported=supported))
lingua_franca.UnsupportedLanguageError: 
Language 'uk' is not yet supported by Lingua Franca. Supported language codes include the following:
da de en es fr hu it nl pt sv

ChanceNCounter: [My fork of this branch](https://github.com/MycroftAI/lingua-franca/compare/refactor/languages_cleanup...ChanceNCounter:refactor/languages_cleanup) is now fully migrated to that function discovery thingy, and seems to be fully functional.

ChanceNCounter: Direct link to the branch:

https://github.com/ChanceNCounter/lingua-franca/tree/refactor/languages_cleanup

It's now much more memory-efficient, assuming it doesn't leak without my noticing. It also contains equivalent Formatter (with tests) and Parser (no tests yet, may not work) objects:

```python3
from lingua_franca.Formatter import Formatter
fmt = Formatter(langs=["en", "es"])

ChanceNCounter:
Closing in favor of #108, which is a superset of this branch and contains all the above commits. Reopen this one if the other is declined, nu?

PR #112, opened 2020-05-03
Fix SyntaxWarning where 'is' is used rather than '=='

PureTryOut:
Without this MR at least Python 3.8 will complain:

```
/usr/lib/python3.8/site-packages/lingua_franca/lang/format_en.py:293: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if abs(num) < 1.0 and (result is "minus " or not result):
/usr/lib/python3.8/site-packages/lingua_franca/lang/format_pt.py:108: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if abs(num) < 1.0 and (result is "menos " or not result):
/usr/lib/python3.8/site-packages/lingua_franca/lang/format_it.py:387: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if abs(num) < 1.0 and (result is "meno " or not result):
/usr/lib/python3.8/site-packages/lingua_franca/lang/format_es.py:197: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if abs(num) < 1.0 and (result is "menos " or not result):
/usr/lib/python3.8/site-packages/lingua_franca/lang/format_fr.py:195: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if abs(num) < 1.0 and (result is "moins " or not result):
/usr/lib/python3.8/site-packages/lingua_franca/lang/parse_en.py:404: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if ordinals and prev_word in string_num_ordinal and val is 1:
```

### DIFF
--- a/lingua_franca/lang/format_en.py
+++ b/lingua_franca/lang/format_en.py
@@ -290,7 +290,7 @@ def pronounce_number_en(num, places=2, short_scale=True, scientific=False,
         return pronounce_number_en(num, places, short_scale, scientific=True)
     # Deal with fractional part
     elif not num == int(num) and places > 0:
-        if abs(num) < 1.0 and (result is "minus " or not result):
+        if abs(num) < 1.0 and (result == "minus " or not result):
             result += "zero"
         result += " point"
         _num_str = str(num)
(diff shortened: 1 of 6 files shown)


ChanceNCounter:
Looks good. I'm gonna tag "don't merge" for a couple days so that we can merge this onto an anticipated refactor, rather than rebasing that huge branch for what would be the third time.

But, otherwise, looks correct at first glance.

PR #115, opened 2020-05-18
Added wheel and pytest as dependencies

adocampo:
Modified README to make sure contributors are installing these tools inside the virtual environment, so global-wide pytest and python-wheel won't interfere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# vscode settings
+.vscode/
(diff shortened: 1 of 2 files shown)


krisgesling:
Hey, great addition to the readme!

My only suggestion is to bump the install of wheel to first in the list of commands as the `python -m pip install .` may use it.

ChanceNCounter:
Thanks as always, @adocampo! I went ahead and moved wheel up, per @krisgesling, and I'm gonna go ahead and merge now. Leaving all 3 commits, as the first 2 do different things and the 3rd would overwrite your authorship.

PR #119, opened 2020-08-06
Small fix to enable german duration parsing

emphasize:
In the current implementation it is not possible (for german clients) to parse german duration phrases without running into exceptions like gras64 described [here](https://github.com/MycroftAI/mycroft-timer/issues/70)

I added the specific lines to parse_de.py and rewrote the regex pattern and the time_units / {unit} handling, which could be adopted across the language spectrum.

In order to take affect **additional edits** in [mycroft-core/mycroft/util/parse.py](https://github.com/MycroftAI/mycroft-core/blob/d6cbccc0cbfbf6b540830aab4c15a32adf67b9bf/mycroft/util/parse.py#L222) have to be made, like 
```
     if lang_code == "de":
        return extract_duration_de(text)
_log_unsupported_language(lang_code, ['en, de'])
```

With the [PR on skill-timer](https://github.com/MycroftAI/mycroft-timer/pull/77) there should (hopefully) be a smooth experience so far as german clients are concerned.

### DIFF
--- a/lingua_franca/lang/parse_de.py
+++ b/lingua_franca/lang/parse_de.py
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from datetime import datetime
+import re
+from datetime import datetime, timedelta
 from dateutil.relativedelta import relativedelta
 from lingua_franca.lang.parse_common import is_numeric, look_for_fractions, \
     extract_numbers_generic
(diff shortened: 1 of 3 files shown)


forslund:
Thanks! Could you add a couple of unittests for the function in [test_parse_de.py](https://github.com/MycroftAI/lingua-franca/blob/master/test/test_parse_de.py)? If you're unfamiliar with the test framework you can use the [English tests](https://github.com/MycroftAI/lingua-franca/blob/master/test/test_parse.py#L194) as a template.

lingua_franca/parse.py
```diff
@@ -200,9 +201,11 @@ def extract_duration(text, lang=None):
         return extract_duration_en(text)
     if lang_code == "cs":
         return extract_duration_cs(text)
+    if lang_code == "de":
+        return extract_duration_cs(text)
```
ChanceNCounter, line 205: Howdy, and thanks for this! Couple of things that, frankly, remind me of myself 😛 

`return extract_duration_de(text)`

lingua_franca/parse.py
```diff
@@ -200,9 +201,11 @@ def extract_duration(text, lang=None):
...
 
     # TODO: extract_duration for other languages
-    _log_unsupported_language(lang_code, ['en','cs'])
+    _log_unsupported_language(lang_code, ['en','cs', 'de'])
```
ChanceNCounter, line 208: This entry should've been there before, but you're adding support for `de`, so unless I need more coffee I don't think we need it anymore.

lingua_franca/lang/parse_de.py
```diff
@@ -81,6 +82,62 @@
...
+        'hours'        : 'stunden',
+        'days'         : 'tage',
+        'weeks'        : 'wochen'
+    }
+
+    #### Einzahl und Mehrzahl
+    pattern = r"(?P<value>\d+(?:\.?\d+)?)(?:\s+|\-){unit}[ne]?"
+
+    #### Einstiegspunkt für Text-zu-Zahlen Konversion
+    #text = _convert_words_to_numbers_de(text)
```
ChanceNCounter, line 125: insert TODO here (many contributors' editors highlight TODOs, which will make this easier to spot when the time comes to integrate this function)

emphasize: Fixed but idk what to make of these

> This entry should've been there before, but you're adding support for de, so unless I need more coffee I don't think we need it anymore.

lingua_franca/parse.py
emphasize: Get the coffee ready, since the master's parse.py

> 
>    if lang_code == "en":
>         return extract_duration_en(text)
>     if lang_code == "cs":
>         return extract_duration_cs(text)
> 
>     # TODO: extract_duration for other languages
>     _log_unsupported_language(lang_code, ['en','cs'])

ChanceNCounter: Ah. Mea culpa, I changed that bit's behavior in a refactor I've been working on. I forgot we were passing the supported list explicitly.

forslund:
@ChanceNCounter can you merge I no longer have my old super-powers :)

PR #124, opened 2020-08-19
reduce runtime for duration parsing

maxbachmann:
This PR reduces the runtime of `extract_duration` by combining the two regex calls.

### DIFF
--- a/lingua_franca/lang/parse_cs.py
+++ b/lingua_franca/lang/parse_cs.py
@@ -633,31 +633,29 @@ def extract_duration_cs(text):
     # Czech inflection for time: minuta,minuty,minut - safe to use minut as pattern
     # For day: den, dny, dnů - short patern not applicable, list all
 
-    time_units_en = {
-        'microseconds': None,
-        'milliseconds': None,
-        'seconds': None,
-        'minutes': None,
-        'hours': None,
-        'days': None,
-        'weeks': None
-        }
+    time_units = {
+        'microseconds': 0,
+        'milliseconds': 0,
+        'seconds': 0,
+        'minutes': 0,
+        'hours': 0,
+        'days': 0,
+        'weeks': 0
+    }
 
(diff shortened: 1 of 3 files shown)


lingua_franca/lang/parse_cs.py
```diff
@@ -633,31 +633,29 @@ def extract_duration_cs(text):
...
     text = _convert_words_to_numbers_cs(text)
 
-    for unit in _TIME_UNITS_CONVERSION:
-        unit_pattern = pattern.format(unit=unit)
-        matches = re.findall(unit_pattern, text)
-        value = sum(map(float, matches))
-        unit_en= _TIME_UNITS_CONVERSION.get(unit) # Find unit in english
-        # Check if is neccesary to write value - handle (days, weeks)
-        if time_units_en[unit_en] is None or time_units_en.get(unit_en) == 0:
-            time_units_en[unit_en] = value # Write value to english unit
```
maxbachmann, line 656: In czech when using multiple durations of the same unit kind (with different declination of the unit name) it was previously only stored the first time, while subsequent instances of the unit kind got ignored. In english and german this is never done, which is why in this PR all values get considered for czech aswell.

ChanceNCounter:
@krisgesling run test?

krisgesling:
Our Travis account got borked recently and we had to re-authenticate so I'm assuming this was leftover from that

krisgesling (approved):
Hey Max, this looks great, thanks for the optimisation across langauges. 

We've been holding PR's until the big refactor is in but it sounds like Chance might be willing to do one more rebase. So I'll mark this as approved and we might get it in for v0.2.x

ChanceNCounter:
I'm sorry it took so long. The review process ended up occurring in phases. Unfortunately, it turned out to be almost impossible to work on the refactor with the codebase changing repeatedly.

But the end of a refactor is always one big rebase, so now's the time. Thanks, Max!

Issue #139, opened 2020-10-30
Czech formatting tests run for all languages


krisgesling:
Just noticed that in the Czech formatting tests, there are a number that loop through all languages eg [test_nice_date](https://github.com/MycroftAI/lingua-franca/blob/85fbc123da564641d053255f5ac5f2dedcb5e8ae/test/test_format_cs.py#L517)

These are already being run by `test_format.py` so could be taken out. However I can also see the value of shifting them from that central location and be run with the other language specific tests. This might assist with the issue of rapidly loading and unloading languages.

Tony763:
Any action needed?

ChanceNCounter:
I haven't gotten a good look yet, but, yeah, ideally remove for loops and run tests just for the one language.

Some of the tests load information from other files, formatting info and such, and these were *all* originally `for` loops in `test_format`. Easy to catch them in a copy/paste.

Tony763:
Yes, with Czech support I did not check these things, I simply took English tests and altered them to Czech. I will dig into it.

ChanceNCounter:
Up to you. Looks like we're (finally) merging the refactor on Monday (Gez didn't wanna do it on what, for him, is Friday afternoon =P)

So you could do it now off of master, and we'll rebase. You could do it now off of the refactor, and that'll be master by this time on Tuesday. Or you could wait until Tuesday =P

Tony763:
Is it possible to make separte test_format_en and test_parse_en? For implementing new languages it would be better to know what should be translated and what is same for all. Other languages are not good for example as they often do not contain all functions.

ChanceNCounter:
Yes, there should have been an issue open for this already. Good catch.

fixed by #144

PR #144, opened 2020-11-02
test_format_cs: remove testing for other languages

Tony763:
Address issue #139
Removed for lops
Run tests only for "cs-cz"
All test passed


ChanceNCounter:
First glance, looks right. Travis seems to be having a bad day, so it might take a while before CI actually finishes.

PR #146, opened 2020-11-02
bug fix #145 #143 #135 #153

### LINKED ISSUE #135: normalize mishandles multi-word numbers
```python3
>>> parse.normalize("four thirty seven")
'4 30 7'
```

### LINKED ISSUE #143: bug in extract datetime - numbers not normalized
"2 minutes" works, "two minutes" does not

this is because numbers are not normalized

this both is and isnt a simple fix, text should just be normalized BUT

- normalize only replaces numbers up to 20
- i added a partial fix to this
- fix was reverted, see discussion here https://github.com/MycroftAI/mycroft-core/issues/1962

this should probably be split into two PRs/discussions, one adding the normalization in extract_datetime, other for fixing normalization

### LINKED ISSUE #145: bug - extract duration lowercases everything
extract_duration should not lower case the text in the remainder text, this might mess up comparisons with original string and simple replacements

### LINKED ISSUE #153: better ordinals handling
when working on #135 i realized we need better control over ordinals flag

- first is non ambiguous 
- second can be the time unit, not a number
- third and above can be a fraction
- things like "half" should not be normalized either, but parsed by extract_number

the same distinction might be useful for extract_number, therefore i propose

ordinals=True -> consider edge cases an integer
ordinals=False -> consider edge cases a fraction
ordinals=None -> ignore ambiguous words

things like 1st, 2nd, 3rd, 4th... are explicit numbers and should always be extracted

things like "whole", "half" or other spoken language quirks should also be optional

JarbasAl:
- case preservation in extract_duration
- extract_datetime number words support
- number replacement in normalize
- adds better support for ordinals

fixes #145
fixes #143 
fixes #135
fixed #153

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -24,7 +24,7 @@ from lingua_franca.lang.common_data_en import _ARTICLES_EN, _NUM_STRING_EN, \
     _NEGATIVES_EN, _SUMS_EN, _MULTIPLIES_LONG_SCALE_EN, \
     _MULTIPLIES_SHORT_SCALE_EN, _FRACTION_MARKER_EN, _DECIMAL_MARKER_EN, \
     _STRING_NUM_EN, _STRING_SHORT_ORDINAL_EN, _STRING_LONG_ORDINAL_EN, \
-    _FRACTION_STRING_EN, _generate_plurals_en
+    _FRACTION_STRING_EN, _generate_plurals_en, _SPOKEN_EXTRA_NUM_EN
 
 import re
 import json
@@ -45,7 +45,6 @@ def _convert_words_to_numbers_en(text, short_scale=True, ordinals=False):
         The original text, with numbers subbed in where appropriate.
 
     """
-    text = text.lower()
     tokens = tokenize(text)
     numbers_to_replace = \
         _extract_numbers_with_text_en(tokens, short_scale, ordinals)
@@ -274,7 +273,7 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
 
     """
     multiplies, string_num_ordinal, string_num_scale = \
-        _initialize_number_data_en(short_scale)
+        _initialize_number_data_en(short_scale, speech=ordinals is not None)
 
     number_words = []  # type: [Token]
     val = False
@@ -287,16 +286,17 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
             next_val = None
             continue
 
-        word = token.word
+        word = token.word.lower()
         if word in _ARTICLES_EN or word in _NEGATIVES_EN:
             number_words.append(token)
             continue
 
-        prev_word = tokens[idx - 1].word if idx > 0 else ""
-        next_word = tokens[idx + 1].word if idx + 1 < len(tokens) else ""
+        prev_word = tokens[idx - 1].word.lower() if idx > 0 else ""
+        next_word = tokens[idx + 1].word.lower() if idx + 1 < len(tokens) else ""
 
         if is_numeric(word[:-2]) and \
-                (word.endswith("st") or word.endswith("nd") or word.endswith("rd") or word.endswith("th")):
+                (word.endswith("st") or word.endswith("nd") or
+                 word.endswith("rd") or word.endswith("th")):
 
             # explicit ordinals, 1st, 2nd, 3rd, 4th.... Nth
             word = word[:-2]
@@ -307,6 +307,8 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 tokens[idx + 1] = Token("", idx)
                 next_word = ""
 
+        # TODO replaces the wall of "and" and "or" with all() or any() as
+        #  appropriate, the whole codebase should be checked for this pattern
         if word not in string_num_scale and \
                 word not in _STRING_NUM_EN and \
                 word not in _SUMS_EN and \
@@ -316,7 +318,8 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 not is_fractional_en(word, short_scale=short_scale) and \
                 not look_for_fractions(word.split('/')):
             words_only = [token.word for token in number_words]
-            if number_words and not all([w in _ARTICLES_EN |
+
+            if number_words and not all([w.lower() in _ARTICLES_EN |
                                          _NEGATIVES_EN for w in words_only]):
                 break
             else:
@@ -329,8 +332,13 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 and prev_word not in _NEGATIVES_EN \
                 and prev_word not in _ARTICLES_EN:
             number_words = [token]
+
         elif prev_word in _SUMS_EN and word in _SUMS_EN:
             number_words = [token]
+        elif ordinals is None and \
+                (word in string_num_ordinal or word in _SPOKEN_EXTRA_NUM_EN):
+            # flagged to ignore this token
+            continue
         else:
             number_words.append(token)
 
@@ -374,12 +382,15 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
 
         # is this a spoken fraction?
         # half cup
-        if val is False:
-            val = is_fractional_en(word, short_scale=short_scale)
+        if val is False and \
+                not (ordinals is None and word in string_num_ordinal):
+            val = is_fractional_en(word, short_scale=short_scale,
+                                   spoken=ordinals is not None)
+
             current_val = val
 
         # 2 fifths
-        if not ordinals:
+        if ordinals is False:
             next_val = is_fractional_en(next_word, short_scale=short_scale)
             if next_val:
                 if not val:
@@ -400,7 +411,7 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 current_val = val
 
         else:
-            if all([
+            if current_val and all([
                 prev_word in _SUMS_EN,
                 word not in _SUMS_EN,
                     word not in multiplies,
@@ -470,8 +481,8 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
 
                 time_to_sum = True
                 for other_token in tokens[idx+1:]:
-                    if other_token.word in multiplies:
-                        if string_num_scale[other_token.word] >= current_val:
+                    if other_token.word.lower() in multiplies:
+                        if string_num_scale[other_token.word.lower()] >= current_val:
                             time_to_sum = False
                         else:
                             continue
@@ -488,14 +499,15 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
     return val, number_words
 
 
-def _initialize_number_data_en(short_scale):
+def _initialize_number_data_en(short_scale, speech=True):
     """
     Generate dictionaries of words to numbers, based on scale.
 
     This is a helper function for _extract_whole_number.
 
     Args:
-        short_scale boolean:
+        short_scale (bool):
+        speech (bool): consider extra words (_SPOKEN_EXTRA_NUM_EN) to be numbers
 
     Returns:
         (set(str), dict(str, number), dict(str, number))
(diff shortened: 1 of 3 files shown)


lingua_franca/lang/parse_en.py
```diff
@@ -622,8 +621,12 @@ def extract_datetime_en(text, anchorDate=None, default_time=None):
...
+        # normalize and lowercase utt  (replaces words with numbers=
+        s = normalize_en(s, remove_articles=False).lower()
+        # restore some decimal markers
+        s = s.replace("0.5", "half").replace("0.25", "quarter")
         # clean unneeded punctuation and capitalization among other things.
-        s = s.lower().replace('?', '').replace('.', '').replace(',', '') \
+        s = s.replace('?', '').replace('.', '').replace(',', '') \
             .replace(' the ', ' ').replace(' a ', ' ').replace(' an ', ' ') \
             .replace("o' clock", "o'clock").replace("o clock", "o'clock") \
             .replace("o ' clock", "o'clock").replace("o 'clock", "o'clock") \
```
ChanceNCounter, line 632: OOS here, but surely this could be a regex looking for (invalid regex) `o*'*clock` (noting here because I just noticed)

lingua_franca/lang/parse_en.py
```diff
@@ -45,7 +45,6 @@ def _convert_words_to_numbers_en(text, short_scale=True, ordinals=False):
         The original text, with numbers subbed in where appropriate.
 
     """
-    text = text.lower()
     tokens = tokenize(text)
```
ChanceNCounter, line 48: Pros and cons vs. `tokens = [token.lower() for token in tokenize(text)]`?

ChanceNCounter (commented):
Just the one question, and an out-of-scope observation that jumped out at me.

lingua_franca/lang/parse_en.py
JarbasAl: tokens are objects, not strings, and they should keep original string for phrase reconstruction

the .lower() operation is only relevant in comparisons and should NOT be propagated to the remainder text

lingua_franca/lang/parse_en.py
JarbasAl: out of scope, will be handled once my datetime rewrite is finished, essentially any changes to extract_datetime are currently a waste of time

test/test_parse.py
```diff
@@ -221,27 +221,52 @@ def test_extract_duration_en(self):
...
+            extractedDate = extractedDate.strftime("%Y-%m-%d %H:%M:%S")
+            return [extractedDate, leftover]
+
+        def testExtract(text, expected_date, expected_leftover):
+            res = extractWithFormat(normalize(text))
+            self.assertEqual(res[0], expected_date, "for=" + text)
+            self.assertEqual(res[1], expected_leftover, "for=" + text)
+
+        testExtract("Set the ambush for half an hour",
+                    "2017-06-27 13:34:00", "set ambush")
```
JarbasAl, line 260: NOTE: remainder is also lowercased here, opening a new issue for this elsewhere

test/test_parse.py
```diff
@@ -697,15 +739,16 @@ def test_numbers(self):
         self.assertEqual(normalize("that's eighteen nineteen twenty"),
                          "that is 18 19 20")
         self.assertEqual(normalize("that's one nineteen twenty two"),
-                         "that is 1 19 20 2")
+                         "that is 1 19 22")
```
JarbasAl, line 742: for these unittests we might want to throw a deprecation warning in case some downstream depends on this, relevant discussion https://github.com/MycroftAI/mycroft-core/issues/1962

test/test_parse.py
JarbasAl: documented here https://github.com/MycroftAI/lingua-franca/issues/147

ChanceNCounter:
Wellp, I was just being obtuse with my only review comment =P

Tests passing, satisfies the issues, that DeprecationWarning is up to you. Looks good to me.

JarbasAl:
id rather not throw the deprecation warning, i consider this a bug, you dont tell your users "we are delaying this fix until next release, work around it somehow"

ChanceNCounter:
We could *raise* a DeprecationError, so at least stuff in the wild breaks... descriptively?

ChanceNCounter:
Suppose that'd require you to put the correct behavior on a flag in the interim, which still isn't great...

test/test_parse.py
```diff
@@ -71,20 +71,107 @@ def test_articles(self):
...
+                                        ordinals=None), False)
+
+        self.assertEqual(extract_number("one third of a cup"), 1.0 / 3.0)
+        self.assertEqual(extract_number("one third of a cup",
+                                        ordinals=True), 3)
+        self.assertEqual(extract_number("one third of a cup",
+                                        ordinals=None), 1)
+
+        # test plurals
+        # NOTE plurals are never considered ordinals, but also not
```
JarbasAl, line 120: @ChanceNCounter @krisgesling does this note make sense?

test/test_parse.py
```diff
@@ -71,20 +71,107 @@ def test_articles(self):
...
+                                        ordinals=True), 5)
+        self.assertEqual(extract_number("2 fifths",
+                                        ordinals=False), 2/5)
+        self.assertEqual(extract_number("2 fifths",
+                                        ordinals=None), 2)
+
+        self.assertEqual(extract_number("Twenty two and Three Fifths"), 22.6)
+
+        # TODO fix me - see issue #152
+        # these currently return 22 instead of 3
```
JarbasAl, line 134: once #152 if fixed this test needs updated, tagging the issue here for reference

test/test_parse.py
```diff
@@ -697,13 +831,15 @@ def test_numbers(self):
...
         self.assertEqual(normalize("that's one hundred"),
-                         "that is 1 hundred")
+                         "that is 100")
         self.assertEqual(normalize("that's one two twenty two"),
-                         "that is 1 2 20 2")
+                         "that is 1 2 22")
         self.assertEqual(normalize("that's one and a half"),
                          "that is 1 and half")
+
+        # TODO fix me
```
JarbasAl, line 842: this test is failing and i could not figure out why yet, this is the main blocker for merging this PR currently, marking as draft until solved

test/test_parse.py
JarbasAl: i dont think we need a warning, if anything dowtreamns is depending on this behavior it is already utterly broken for n > 20 in general... we would have received bug reports already.

lingua_franca/lang/parse_en.py
```diff
@@ -400,7 +409,7 @@ def _extract_whole_number_with_text_en(tokens, short_scale, ordinals):
                 current_val = val
 
         else:
-            if all([
+            if current_val and all([
```
JarbasAl, line 412: this is an unrelated minor bug fix, during testing i ran into a case where current_val was None and int comparison inside all() would throw an exception

JarbasAl:
eventually ordinals flag in extract number should become an Enum, for backwards compatibility this should not be changed yet but only in next major version

test/test_parse.py
```diff
@@ -71,20 +71,105 @@ def test_articles(self):
...
+        self.assertEqual(extract_number("Twenty two and Three Fifths"), 22.6)
+
+        # TODO fix me - see issue #152
+        # these currently return 22 instead of 3
+        #self.assertEqual(extract_number("Twenty two and Three Fifths",
+        #                                ordinals=True), 3)
+
+        # test multiple ambiguous
+        self.assertEqual(extract_number("sixth third"),
+                         1 / 6 / 3)
```
krisgesling, line 140: I don't think this one makes sense. 
"Sixth third" would probably refer to "the sixth instance of a third" which I guess means there's `2` all up.

test/test_parse.py
ChanceNCounter: Since you have to extract ordinals explicitly, and there's no such ordinal as The Three Fifth, it's probably sensible to avoid returning fractions in all cases.

On the other hand, it might make sense to call them ordinals, on the assumption that `extract_number("the thirds", ordinals=True)` will always be an input failure somewhere upstairs, like STT mishearing "the third."

test/test_parse.py
ChanceNCounter: Agreed. I dunno what should be returned here, but I don't think it should be cumulative.

JarbasAl: this one has been around for ages, i just changed it to a different place so all related tests are grouped together

i dont think it makes a lot of sense either, but i'm declaring it out of scope since it isnt touched by this PR

test/test_parse.py
```diff
@@ -71,20 +71,105 @@ def test_articles(self):
...
+        self.assertEqual(extract_number("this is the third test"),
+                         1.0 / 3.0)
+        self.assertEqual(extract_number("this is the third test",
+                                        ordinals=None), False)
+
+        self.assertEqual(extract_number("one third of a cup"), 1.0 / 3.0)
+        self.assertEqual(extract_number("one third of a cup",
+                                        ordinals=True), 3)
+        self.assertEqual(extract_number("one third of a cup",
+                                        ordinals=None), 1)
```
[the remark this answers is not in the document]
JarbasAl: at some point i would like to replace this with an Enum, i did it this way for backwards compat only. but imho None should be the default in a next major release

test/test_parse.py
```diff
@@ -71,20 +71,105 @@ def test_articles(self):
...
+        self.assertEqual(extract_number("remind me in a second",
+                                        ordinals=False), False)
+        self.assertEqual(extract_number("remind me in a second",
+                                        ordinals=None), False)
+
+        # test ambiguous ordinal/fractional
         self.assertEqual(extract_number("this is the third test",
                                         ordinals=True), 3.0)
+        self.assertEqual(extract_number("this is the third test"),
+                         1.0 / 3.0)
```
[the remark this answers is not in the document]
JarbasAl: makes sense, will change :+1:

test/test_parse.py
JarbasAl: i was unsure of this one, for consistency i thought it made sense that ordinals=True would not parse fractions at all, the same way ordinals=False would only parse fractions.  But since we support explict ordinals (1st, 2nd, 3rd, 4th...) i'm not sure....

i guess its a matter of how explicit is this? what are the possible sources of ambiguity?

"the thirds" could very well mean 3 instead of 1/3 , so i don't think this should be considered an explicit fraction

JarbasAl: i wonder if when ordinals=True we should consider the plural form an integer?

test/test_parse.py
```diff
@@ -71,20 +71,105 @@ def test_articles(self):
...
+
+        # TODO fix me - see issue #152
+        # these currently return 22 instead of 3
+        #self.assertEqual(extract_number("Twenty two and Three Fifths",
+        #                                ordinals=True), 3)
+
+        # test multiple ambiguous
+        self.assertEqual(extract_number("sixth third"),
+                         1 / 6 / 3)
+        self.assertEqual(extract_number("sixth third", ordinals=True),
```
JarbasAl, line 141: this one might hint at a deeper issue, since #152 should be breaking this test but isn't

test/test_parse.py
```diff
@@ -671,6 +756,54 @@ def testExtract(text, expected_date, expected_leftover):
...
+        self.assertEqual(normalize("remind me to do something at twenty to two"),
+                         "remind me to do something at 20 to 2")
+
+        # TODO imperfect test, maybe should return 'my favorite numbers are 20 2',
+        #  let is pass for now since this is likely a STT issue if ever
+        #  encountered in the wild and is somewhat ambiguous, if this was
+        #  spoken by a human the result is what we expect, if in written form
+        #  it is ambiguous but could mean separate numbers
+        self.assertEqual(normalize('my favorite numbers are twenty 2'),
+                         'my favorite numbers are 22')
```
krisgesling, line 775: may also be worth adding in where both numbers are digits 
> "my favorite numbers are 20 2"

Which I would expect would remain unchanged.

JarbasAl: added a test for this, but it is also changed. This shares logic with extract_number and will be hard to fix without a bigger refactor, if you are ok with that i would like to leave it for a follow up PR

test/test_parse.py
JarbasAl: added a code comment for this one

JarbasAl:
whats blocking this?

ChanceNCounter:
Next in line after Catalan, let this block release IMO. Shouldn't be too bad. The will-be merge conflicts look like they should be click-fixes.

ChanceNCounter:
squashed and pushed. speak soonish or forever hold your peace 🚀

PR #148, opened 2020-11-03
explicitly add files to source dist package

krisgesling:
Key files are not being included in the source distribution. This adds a `MANIFEST.in` to explicitly include them.

We hit the same issue with the Messagebus Client recently and installations will fail without them present.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include LICENSE
 include readme.md
 include requirements.txt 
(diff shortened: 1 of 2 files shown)


ChanceNCounter:
I'm bad at setuptools. Buck passed.

JarbasAl:
legit

JarbasAl:
looks good to me :+1:

PR #151, opened 2020-11-03
Fix/deprwarn bugs

ChanceNCounter:
Several problems slipped through when 0.3.0 was altered to deprecate, rather than remove, `lang=None` and etc. as valid parameters.

The default value of the lang param, in several places, needs to change to an empty string, rather than None. This can be reverted once the deprecation is complete, and it probably should be, because future development will be much more pleasant when `if not lang` behaves consistently.

Also, I managed to leave a `str().append()` in there, I dunno how the suites missed that.

### DIFF
--- a/lingua_franca/format.py
+++ b/lingua_franca/format.py
@@ -41,7 +41,7 @@ _REGISTERED_FUNCTIONS = ("nice_number",
 populate_localized_function_dict("format", langs=get_active_langs())
 
 
-def _translate_word(name, lang=None):
+def _translate_word(name, lang=''):
     """ Helper to get word translations
 
     Args:
@@ -53,6 +53,8 @@ def _translate_word(name, lang=None):
     """
     from lingua_franca.internal import resolve_resource_file
     if not lang:
+        if lang == None:
+            warn(NoneLangWarning)
         lang = get_default_loc()
 
     lang_code = lang if is_supported_full_lang(lang) else \
@@ -241,7 +243,7 @@ date_time_format = DateTimeFormat(os.path.join(os.path.dirname(__file__),
 
 
 @localized_function(run_own_code_on=[UnsupportedLanguageError])
-def nice_number(number, lang=None, speech=True, denominators=None):
+def nice_number(number, lang='', speech=True, denominators=None):
     """Format a float to human readable functions
 
     This function formats a float to human understandable functions. Like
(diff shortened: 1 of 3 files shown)


lingua_franca/format.py
```diff
@@ -385,7 +387,8 @@ def nice_duration(duration, lang=None, speech=True):
         str: timespan as a string
     """
     if not lang:
-        warn(NoneLangWarning)
+        if lang == None:
```
JarbasAl, line 390: nitpick, should be ```if lang is None```

https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

ChanceNCounter: aww crud that got missed in a push just now

lingua_franca/internal.py
```diff
@@ -37,6 +37,10 @@
 # commit 35efd0661a178e82f6745ad17e10e607c0d83472 for the "proper" state
 # of affairs, raising the errors below instead of deprecation warnings
 
+# For ease of development, in order to be able to type `if not lang:`
```
JarbasAl, line 40: can you explain better? if ```lang is None```,  then ```if lang:``` is falsish

you can do ```if not lang:``` freely in all cases

```
lang = None
if not lang:
    print("not lang None")
lang = ""
if not lang:
    print("not lang empty")
lang = False
if not lang:
    print("not lang Bool")
```

ChanceNCounter: Well, this change is necessary because the deprecation warnings currently rely on `lang` being `None` only when the parameter is actually passed: `nice_duration(60, lang=None)`

We had been using `lang=None` as the default param for all these functions, and I'd like to continue doing so, but the interaction between the two causes the DeprecationWarnings to come up *by default*, when no language is passed.

Hence, right now, and until the None-lang deprecation happens, this module draws a distinction between `lang is None` and `lang == ''`

ChanceNCounter: In fact, this makes most of the test suite warnings go away.

JarbasAl: i get that, but you said 

	in order to be able to type `if not lang:`

which i maybe misunderstood as you saying "if lang == "" this will fail"

JarbasAl: what i am saying is that i disagree with this comment, i don't see a need to make the default None again at a later date

JarbasAl:
LGTM, merging

ChanceNCounter:
Travis is finally done, got the 👍, pushing button

PR #154, opened 2020-11-04
Add a Github Action for pytest

krisgesling:
#### Description
This replaces Travis CI with Github Actions for the purpose of running pytest across multiple versions of Python.

Travis has become extremely slow and will require migration to their new .com service in the very near future.

#### Type of PR
- [ ] Bugfix
- [ ] Feature implementation
- [x] Test improvements

#### Testing
Check the Action output


ChanceNCounter:
lolwut at a test failing under 3.5 that Travis passes... this warrants investigation

krisgesling:
I noticed that only one of those tests didn't explicitly pass a lang argument but this should be optional right?

ChanceNCounter:
should be, yeah, unless that spaghetti-default thing comes back, but I pray that bug is restricted to multiple languages within a file

krisgesling:
I've updated this PR to actually replace Travis. 

Github Actions seem to be pretty reliable and the Travis-CI.org will be deprecated at some point in the next 6 weeks.

ChanceNCounter:
Well, I haven't bothered to learn the syntax, but all the commands look right, and the tests run =P

I think we should consider adding coverage to this, as long as we've moved to a more free-form CI thingy, but that's another issue for another day.

ChanceNCounter (approved):

Issue #128, opened 2020-09-11
Move runtime config (default language, localization settings, etc) to a stateful object


ChanceNCounter:
Hear me out 😛 

The module is (or will soon be) thoroughly respectful of a given language's default rules (for instance, short vs. long scale numbers.) We've also got some, and will obviously extent, support for full localization (en_US, en_GB, etc.)

But what if the user only wants *some* of their language's "default" rules? It's easy to look at this as an implementation detail, but the values in question are effectively default *parameters* to our functions. If we read this information onto a stateful object, the implementation (such as Mycroft) could nitpick on the fly from end-user config files.

For instance, Italian uses the long scale, but it's entirely possible that an Italian speaker will prefer the short scale, especially if they're working bilingually. They want to override just the one default parameter. Should it really be the implementation's problem to repeatedly and explicitly pass that information to each parser and formatter? It's particularly problematic for skill authors, who are invoking LF functions by lang code, and relying on Mycroft to know the details, which then relies on LF.

krisgesling:
Short scale is only for insecure rich people who need to feel even richer than they already are. It should be abolished! 
_\*\*Ducks from all the Americans**_

But back to the real issue. Would an example be `en_AU` where technically the long scale is officially correct, but many people use the short scale because of TV (or something)? 

Would this need to be changed on the fly or only when a language is loaded? Presumably this would override the normal default for each method, but if it's intentionally passed in a method call then the Skill (or code) that called it can override the user defined config? eg:
1. lang default
2. user config
3. method call
?

ChanceNCounter:
`en_AU` and the number scale is a perfect example!

I think the goal is to make it changeable at runtime, but we don't need to get fancy with the data structures. Just provide some getters and setters so that the implementation can fix things up from *its* config files.

`lingua_franca.config.set_number_scale('en_AU', 'short')`

In practice, I figure (for example) that Mycroft would call those functions when it imports LF, shortly after loading languages. Some kind of for loop over the relevant config values.

I saw another FOSS assistant project, with some familiar faces :heart:, eyeing Lingua Franca. I don't know if they actually adopted us, but, if so, their config files will surely look different from Mycroft's; as long as we stick to get/sets, that won't matter to LF!

ChanceNCounter:
[...]

PR #158, opened 2020-11-24
config.load_langs_on_demand: permits on-the-fly loading/unloading, to facilitate proc-gen use cases

ChanceNCounter:
This branch (squash pls) adds `config.py` (eventually #128) with one variable: `load_langs_on_demand`.

When this variable is `True`, the localized function caller will - if and *only* if a language is passed which is not loaded - load the language, call the function, and unload the language.

At the moment, this is pretty fast, though it might need to be revisited if some languages become truly bloated with resource files.

This is only desirable in certain use cases, but, for instance, the absence of this flag was an obstacle for @JarbasAl's modular `text2speech` package, which doesn't always know the language it wants ahead of time.

I've tested this PR against that module, and it solves the problem. I want to stress that use cases like Mycroft, which prefer to utilize at most one engine per language, will *not* want to set this flag.

### DIFF
--- a/lingua_franca/__init__.py
+++ b/lingua_franca/__init__.py
@@ -2,3 +2,5 @@ from .internal import get_default_lang, set_default_lang, get_default_loc, \
     get_active_langs, _set_active_langs, get_primary_lang_code, \
     get_full_lang_code, resolve_resource_file, load_language, \
     load_languages, unload_language, unload_languages, get_supported_langs
+
+from . import config
--- a/lingua_franca/__init__.py
+++ b/lingua_franca/__init__.py
@@ -3,4 +3,4 @@ from .internal import get_default_lang, set_default_lang, get_default_loc, \
     get_full_lang_code, resolve_resource_file, load_language, \
     load_languages, unload_language, unload_languages, get_supported_langs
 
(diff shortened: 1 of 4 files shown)


JarbasAl (approved):
I approve of this PR, as discussed in chat with @ChanceNCounter sometimes we will not know what language will be needed, and in these cases the lang parameter will be passed along. 

Since the user is explicitly requesting a language, he wants it loaded for sure, however as @ChanceNCounter pointed out this would bloat memory usage, eg, `normalize(utt, lang="pt") ` for a single utterance would keep pt language data in memory, even if pt is never used again. Unloading makes perfect sense. It's up for end users to explicitly load a language if they want the performance benefits (speed) from not having to load everytime. 

At same time this avoids the need for a pattern of calling load_language, func, unload_language every time we use a function

krisgesling:
Yeah sounds like a good change, agree that the intent is there if the lang param is passed. 

If Jarbas has reviewed the code I'm a +1

My only question is whether we can we add a test to cover this use case?

ChanceNCounter:
Yeah, that's a dumb oversight! On it.

PR #161, opened 2020-11-29
Add Catalan support to Lingua Franca

jmontane:
#### Description
Adds Catalan support to Lingua Franca

#### Type of PR
- [ x] Feature implementation

#### Testing
This PR passes all created tests for Catalan.


ChanceNCounter:
Don't worry about tests failing in Py3.5. We'll be removing support pretty soon. GitHub probably won't run the other tests after 3.5 fails; if this is 100% ready before 3.5 is removed, we can run pytest manually on our machines before merging.

Hence, please hit that magic "request review" button when this is good to go, as we won't be able to go by the checkmark.

ChanceNCounter:
Tests confirmed passing in Py3.8 and Py3.9

lingua_franca/internal.py
```diff
@@ -730,3 +731,46 @@ def resolve_resource_file(res_name, data_dir=None):
...
+                                     "exist for {v}".format(v=key))
+            return func(*args, **kwargs)
+
+        return call_function
+
+    try:
+        return lang_variant_function_decorator
+    except NotImplementedError as e:
+        warn(str(e))
+        return
```
ChanceNCounter, line 775: Does this need to fall back on a default, and, if not, should it crash rather than warning?

ChanceNCounter (changes requested):
Not speaking Catalan, I can't judge the correctness of the output, but this localization compares favorably with the existing Iberian localizations, and I certainly know some of those words!

The new wrapper still feels a little strange, but, if it works, it works. I'd like to see a couple of tests for it, perhaps in test_internal.py, but that and the possible fallback result (see review) are all I've got.

lingua_franca/internal.py
ChanceNCounter: Setting aside for this PR.

ChanceNCounter (approved):

PR #164, opened 2020-12-11
drop support for python 3.5 (EOL)

ChanceNCounter:
We held out almost two whole months! 3.5 is donezo, though, and the supported versions contain some delicious syntactic sugar. To aid with the Catalan localization, a new wrapper has been introduced which takes advantage of that sugar. Furthermore, the 0.3 rebase *attempted* to utilize the newer `ModuleNotFoundError`, but 3.5 didn't have it.

Goodbye, 3.5!

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


JarbasAl:
looks good to me, dont think you forgot any deprecation warning

krisgesling (approved):
Yep, looks good, nice and simple.

PR #172, opened 2021-01-07
More it-it long form number tests

krisgesling:
Add new pronounce_number tests using the default params which sounds like it should default to long form.

Note: This does not address failing tests.

### DIFF
--- a/test/test_format_it.py
+++ b/test/test_format_it.py
@@ -181,6 +181,31 @@ class TestPronounceNumber(unittest.TestCase):
                          "ottocento novantaseimila seicento trentanove"
                          " milioni, seicento trentunomila, ottocento"
                          " novantatre")
+        self.assertEqual(pronounce_number(1000000000, lang="it"),
+                         "Un miliardo")
+        self.assertEqual(pronounce_number(1000000001, lang="it"),
+                         "Un miliardo, uno")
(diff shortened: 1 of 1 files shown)


JarbasAl:
can this please use scientific notation? counting zeros to validate long/short scale is very error prone... with scientific notation we just look at the exponent and can immediately judge if things are working like they should. Since these tests are all about the number of zeros i don't feel too bad requesting such a (code wise) inconsequential thing

test/test_format_it.py
```diff
@@ -181,6 +181,31 @@ def test_convert_hundreds(self):
...
                          " milioni, seicento trentunomila, ottocento"
                          " novantatre")
+        self.assertEqual(pronounce_number(1000000000, lang="it"),
+                         "Un miliardo")
+        self.assertEqual(pronounce_number(1000000001, lang="it"),
+                         "Un miliardo, uno")
+        self.assertEqual(pronounce_number(1000000050, lang="it"),
+                         "Un miliardo, cinquanta")
+        self.assertEqual(pronounce_number(1000123456, lang="it"),
+                         "Un miliardo, cento ventri tre mila, "
```
Stypox, line 191:
```suggestion
                         "Un miliardo, cento venti tre mila, "
```

ChanceNCounter:
Bump: change to sci

PR #175, opened 2021-01-28
fix extract_datetime offset-aware/naive bug

emphasize:
#### Description

Few things done here:
1. Fix Offset aware/naive bug stemming from the comparison of extractedDate and temp [here](https://github.com/MycroftAI/lingua-franca/blob/d30424e19d5ac641f1ccd1d4d91f9d7fec2e8a31/lingua_franca/lang/parse_de.py#L838)
by importing default_timezone() and replacing temps tzinfo
(this bug should throw exeptions across the LF spectrum, if hasYear is empty)

2. Adding the possibility to parse utterances as such "5 June 20:00" - to set a reminder or else (prior 20:00 would be seen as a year date and not parsing 20:00 at all)
- the year has to be > 2000 to be recognized as such

3. remove _de_numbers from parse_de since it's imoprted from parse_common_de

#### Type of PR
If your PR fits more than one category, there is a high chance you should submit more than one PR. Please consider this carefully before opening the PR.
_Either delete those that do not apply, or add an x between the square brackets like so: `- [x]`_
- [x] Bugfix
- [x] Feature implementation
- [ ] Test improvements

#### CLA
👍

### DIFF
--- a/lingua_franca/lang/parse_de.py
+++ b/lingua_franca/lang/parse_de.py
@@ -20,64 +20,7 @@ from lingua_franca.lang.parse_common import is_numeric, look_for_fractions, \
     extract_numbers_generic, Normalizer
 from lingua_franca.lang.common_data_de import _DE_NUMBERS
 from lingua_franca.lang.format_de import pronounce_number_de
-
-de_numbers = {
-    'null': 0,
-    'ein': 1,
-    'eins': 1,
-    'eine': 1,
-    'einer': 1,
-    'einem': 1,
-    'einen': 1,
-    'eines': 1,
-    'zwei': 2,
-    'drei': 3,
-    'vier': 4,
-    'fünf': 5,
-    'sechs': 6,
-    'sieben': 7,
-    'acht': 8,
-    'neun': 9,
-    'zehn': 10,
-    'elf': 11,
-    'zwölf': 12,
-    'dreizehn': 13,
-    'vierzehn': 14,
-    'fünfzehn': 15,
-    'sechzehn': 16,
-    'siebzehn': 17,
-    'achtzehn': 18,
-    'neunzehn': 19,
-    'zwanzig': 20,
-    'einundzwanzig': 21,
-    'zweiundzwanzig': 22,
-    'dreiundzwanzig': 23,
-    'vierundzwanzig': 24,
-    'fünfundzwanzig': 25,
-    'sechsundzwanzig': 26,
-    'siebenundzwanzig': 27,
-    'achtundzwanzig': 28,
-    'neunundzwanzig': 29,
-    'dreißig': 30,
-    'einunddreißig': 31,
-    'vierzig': 40,
-    'fünfzig': 50,
-    'sechzig': 60,
-    'siebzig': 70,
-    'achtzig': 80,
-    'neunzig': 90,
-    'hundert': 100,
-    'zweihundert': 200,
-    'dreihundert': 300,
-    'vierhundert': 400,
-    'fünfhundert': 500,
-    'sechshundert': 600,
-    'siebenhundert': 700,
-    'achthundert': 800,
-    'neunhundert': 900,
-    'tausend': 1000,
-    'million': 1000000
-}
+from mycroft.util.time import default_timezone
 
 # TODO: short_scale and ordinals don't do anything here.
 # The parameters are present in the function signature for API compatibility
@@ -405,7 +348,8 @@ def extract_datetime_de(text, anchorDate=None, default_time=None):
                 dayOffset -= 7
                 used += 1
                 start -= 1
-                # parse 15 of July, June 20th, Feb 18, 19 of February
+
+        # parse 15 Mai, Mai der 20ste, Dez 18
         elif word in months or word in monthsShort and not fromFlag:
             try:
                 m = months.index(word)
@@ -413,10 +357,13 @@ def extract_datetime_de(text, anchorDate=None, default_time=None):
                 m = monthsShort.index(word)
             used += 1
             datestr = months[m]
+            #commonly spoken : 15(.=gets replaced) Mai <Year>/<time>
             if wordPrev and (wordPrev[0].isdigit() or
-                             (wordPrev == "of" and wordPrevPrev[0].isdigit())):
-                if wordPrev == "of" and wordPrevPrev[0].isdigit():
-                    datestr += " " + words[idx - 2]
+                             (((wordNext == "der") or (wordNext == "den")) and
+                             wordNextNext[0].isdigit())):
+                #Mai der fünfte(5)
+                if ((wordNext == "der") or (wordNext == "den")) and wordNextNext[0].isdigit():
+                    datestr += " " + words[idx + 2]
                     used += 1
                     start -= 1
                 else:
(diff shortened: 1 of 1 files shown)


emphasize:
Hm.. probably a bad idea to restrict year to >2000.
I will replace it with >60 since this this would exclude minute (and ofc hour) time data

emphasize:
If this is a bug across the spectrum i will add specific lines to all parsers

lingua_franca/lang/parse_de.py
```diff
@@ -20,64 +20,7 @@
...
-    'vierhundert': 400,
-    'fünfhundert': 500,
-    'sechshundert': 600,
-    'siebenhundert': 700,
-    'achthundert': 800,
-    'neunhundert': 900,
-    'tausend': 1000,
-    'million': 1000000
-}
+from mycroft.util.time import default_timezone
```
JarbasAl, line 23: should not import from mycroft, this is a standalone package

emphasize: hm.. would something like
`temp = datetime.strptime(datestr, "%B %d").replace(tzinfo=extractedDate.tzinfo)`
work?

emphasize: nvm, works. i'll update it

ChanceNCounter: It's also forgivable to override things downstream. See MycroftAI/mycroft-core#2772 which contains a number of little things. Some of those are temporary, some are for backward compatibility.

(Feel free to PR to the PR, via that branch on my fork. It feels weird to PR to a PR, but that's the magic of git!)

ChanceNCounter:
(6 weeks later...)

On reflection, it still seems like a bad idea to restrict year to > 60, because we almost certainly want to support utterances like, "What happened on June 13, '27?"

emphasize:
In german spoken language you normally don't use this kind of abbrevation. As in written language.

Another thing that i just saw. I don't know how the others handle `datestr`. But ours wouldn't work no matter what. (the relevant parts are broken out and python highlighted). The temp datetime object is only created with month and day, which is a problem if `datestr` contains the year.

Back to the abbrevation, a `datestr` like 18. Januar '20 would be also greeted with an exception parsed like this.

So, this has to be adjusted one way or the other. Parsing with `'` as indicator would be easy yet 
i don't think the tts services send it back seen as a year date. 

```
extractedDate = dateNow
extractedDate = extractedDate.replace(microsecond=0,
                                          second=0,
                                          minute=0,
                                          hour=0)

if datestr != "":
    en_months = ['january', 'february', 'march', 'april', 'may', 'june',
               'july', 'august', 'september', 'october', 'november',
               'december']
    en_monthsShort = ['jan', 'feb', 'mar', 'apr', 'may', 'june', 'july',
                      'aug',
                      'sept', 'oct', 'nov', 'dec']
    for idx, en_month in enumerate(en_months):
        datestr = datestr.replace(months[idx], en_month)
    for idx, en_month in enumerate(en_monthsShort):
        datestr = datestr.replace(monthsShort[idx], en_month)
``` 
```python
    temp = datetime.strptime(datestr, "%B %d")
```
        if not hasYear:
            temp = temp.replace(year=extractedDate.year)
            if extractedDate < temp:
                extractedDate = extractedDate.replace(year=int(currentYear),
                                                      month=int(
                                                          temp.strftime(
                                                              "%m")),
                                                      day=int(temp.strftime(
                                                          "%d")))
            else:
                extractedDate = extractedDate.replace(
                    year=int(currentYear) + 1,
                    month=int(temp.strftime("%m")),
                    day=int(temp.strftime("%d")))
    else:
        extractedDate = extractedDate.replace(
            year=int(temp.strftime("%Y")),
            month=int(temp.strftime("%m")),
            day=int(temp.strftime("%d")))
```

**Comment by ChanceNCounter** [comment]
Nope, STT will just send it back as two consecutive numbers, along the lines of "18 Januar 20" (hence my concern, though I guess not if it doesn't apply to German.)

**Comment by emphasize** [comment]
Hm.. Just skimmed through parse_en and found no indicator that parsing a year like '20 is even possible. Maybe i'm missing something.

**Comment by ChanceNCounter** [comment]
Oh, it absolutely doesn't work at the moment. Jarbas has a rewrite roadmapped, and there's (just over the past couple days) been some discussion about TZ management that will play into it. So will the existence of `lingua_franca.config`

Nevertheless, at least in languages that intend to support that format - which is *extremely* common in English, but I obviously can't speak to German! - it seems unwise to place restrictions on years (except 0.)

**Comment by emphasize** [comment]
All this wouldn't be necessary if 
* STT not sending back some weird variations of 10:30 and on the other hand
* not only the first char is checked to validate a number

**Comment by ChanceNCounter** [comment]
That we can't control STT is an ongoing challenge. The module also transitioned slowly into supporting "written" input, which is beginning to pay dividends when STT returns weird variations on stuff like that =P

It'll only get weirder as more STT engines proliferate.

Things like this play back into the normalizer, as well, which *should* be able to sanitize *most* of these edge cases.

Even still, with reliably normalized input, we'd be parsing something like "10 12" in a datetime extractor. This could be 10:12, Oct. 12, or Dec. 10. It's tricky business. This is both the upside and the downside of algorithmic vs. ML parsers. We can find rules, bake in edge cases, and that's that, but disambiguation is hard.

**Comment by emphasize** [comment]
A variation could be just to check ":"
```Python
if if wordnext and wordnext[0].isdigit() and not ":" in wordnext:
    datestr += " " + wordNext
    used += 1
    hasYear = True                         
else:
    hasYear = False
```   

yet, as i said, this is shakey with overwhelming misses.

Nah, it is recognized from Google STT since i call "... 10 uhr 10", yet it randomly returns _10 uhr 10 uhr_, _10.10 uhr_ or _10:10 uhr_ with the infamous "dreißig" bug (13 uhr 30 -> _13 uhr dreißig_; not with 10,20,40,...)

emphasize:
oh boy, oh boy. with this german parsers you stumble from brick to brick.

First of changed back 'temp' tz addition back due to #180

With the addition of the german normalizer another unrecognized problem emerged.
In the former approach (directly changed words in normalzer_de instead of loading a json)
`
'ein': 1,
'eins': 1,
'eine': 1,
'einer': 1,
'einem': 1,
'einen': 1,
'eines': 1,
`

is changed wholesale.
Resulting in spoken sentences like "I have one(1) heck of a day" instead of "I have a heck of a day" (Maybe workable in english, yet isn't in german).

Will add these lines to the normalizer.json, yet this has to be treated in specific parsers with "ein",... kicked out of the normalization

JarbasAl:
can you please add some unittests for what this is supposed to be fixing with the datetime changes?

if some other native speaker could double check this it would be great, but its quite simple so let's not block merging because of that

i think this looks good in general, but need to test properly before hitting the green button

glad to see more normalizers being migrated to the new json mechanism

lingua_franca/lang/parse_de.py
```diff
@@ -833,6 +778,7 @@ def date_found():
             datestr = datestr.replace(monthsShort[idx], en_month)
 
         temp = datetime.strptime(datestr, "%B %d")
+        extractedDate = extractedDate.replace(hour=0, minute=0, second=0)
         if not hasYear:
             temp = temp.replace(year=extractedDate.year)
```
JarbasAl, line 783: this might need changing depending on #180

emphasize: i thought so, yet i don't found a change in other parsers.
`
else:
   #ignore the current HH:MM:SS if relative using days or greater
   if hrOffset == 0 and minOffset == 0 and secOffset == 0:
      extractedDate = extractedDate.replace(hour=0, minute=0, second=0)
`
in case of extract_datetime('today'). If coupled with a to_utc() (problem with weatherskill atm) this is causing possible dateime jumps.
wouldn't do harm if this is replaced also.

JarbasAl: the issue im anticipating is the `temp` datetime being a naive datetime, it will cause an issue with timezones, in #180 you can see i changed that step in every language to add the timezone to the `temp` datetime

emphasize:
> can you please add some unittests for what this is supposed to be fixing with the datetime changes?

wouldn't call this part a fix, more of a convenience addition.
For something like "set a reminder on the 5th of july 19:00 o'clock"
Basically if a day of month is given it would expect a year after july and treat 19:00 as such since
[wordNext[0].isdigit()](https://github.com/MycroftAI/lingua-franca/blob/eee95702016b4013b0d81dc74da98cd2d2f53358/lingua_franca/lang/parse_en.py#L874)

I'll add a test

PR #179, opened 2021-02-28
fix #178: don't install tests, just include

### LINKED ISSUE #178: [openSUSE Tumbleweed] Found test in site-packages/test/ when building from source.
Hello,

I am trying to package this module for openSUSE. I found a file that went to

`/usr/lib/python3.8/site-packages/test/`

which it is not allow in SUSE,  I think they prefer to be in `site-packages/lingua_franca/test/`

I see that test folder was include in setup.py `packages=['test', 'lingua_franca', 'lingua_franca.lang'],`.
Thanks.

ChanceNCounter:
I've removed the `tests` directory from `setup.py`, and included `test/ *.py` via `MANIFEST.in`, which is intended to ensure that tests are included with source distributions, but are not installed to the user's system.

I'm not 100% sure I've done this right, but when I make the sdist and unpack it, both `setup.py test` and `pytest test/` succeed as expected.

(see #178)

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include LICENSE
 include readme.md
-include requirements.txt 
+include requirements.txt
+recursive-include test *.py
 recursive-include lingua_franca/res *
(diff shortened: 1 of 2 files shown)


krisgesling (approved):
Yep not sure why we had test defined as it's own package. Can be safely removed, but nice touch keeping it in the source installs.

PR #186, opened 2021-04-04
Bump version: 0.4.0rc1

ChanceNCounter:


ChanceNCounter:
The release is not dependent on #185 , but if it can be handled quickly, it may be worth waiting for that and bumping to rc2. That way, the current, dummy implementation of config, where `load_langs_on_demand` is just a variable, will never have been released to pip, and will only affect the handful of known downstreams running `master`.

PR #190, opened 2021-04-27
Bump version to v0.4.1

krisgesling:
The LF version in `setup.py` was still listed as a release candidate when the `release/v0.4.0 tag` was made. Unfortunately this means that PyPI would consider it v0.4.0rc1 rather than the full 0.4.0. 

This PR bumps the version up a patch number so we can re-tag and have matching versions for PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ with open("readme.md", "r") as fh:
 
 setup(
     name='lingua_franca',
-    version='0.4.0rc1',
+    version='0.4.1',
     packages=['lingua_franca', 'lingua_franca.lang'],
     url='https://github.com/MycroftAI/lingua-franca',
     license='Apache2.0',


ChanceNCounter:
Balls.

ChanceNCounter (approved):

ChanceNCounter:
Alright. While @krisgesling was out of town, @JarbasAl and I discussed and decided to get 0.4.0 tagged before merging Farsi, just in case. New language, big change, possible "oopsies" afterward.

Hence, this is probably the best time for me to have missed that... vagaries of using `pip install` for testing.

Hold the bump for Farsi? I'll rebase that PR this afternoon US time.

krisgesling:
Yep can do, I'm going to work on the mycroft-core PR today based on the current state of LF. So assuming the Farsi PR only adds localized functions this should be good to go as soon as we release v0.4.1 to pypi

ChanceNCounter:
The Mycroft PR has not yet been bumped to 0.4.0, but I believe you're right about the absence of conflicts (indeed, from 0.3 to 0.4, let alone 0.4.0 to .1)

PR #191, opened 2021-05-05
Clean up of public docstrings

krisgesling:
#### Description
This is a quick clean up of public docstrings in the primary parse.py and format.py files.

- uses a consistent description for the `lang` parameter
- adds some missing arguments
- add type hints for `fuzzy_match` where the input args are self-explanatory
- autopep8 formatting

#### Type of PR
- [ ] Bugfix
- [ ] Feature implementation
- [x] Documentation improvements
- [ ] Test improvements

#### Testing
Easiest is eyeballs here I think.

### DIFF
--- a/lingua_franca/format.py
+++ b/lingua_franca/format.py
@@ -45,7 +45,8 @@ def _translate_word(name, lang=''):
 
     Args:
         name (str): Word name. Returned as the default value if not translated
-        lang (str): Language code, e.g. "en-us"
+        lang (str, optional): an optional BCP-47 language code, if omitted
+                              the default language will be used.
 
     Returns:
         str: translated version of resource name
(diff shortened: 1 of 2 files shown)


lingua_franca/format.py
```diff
@@ -290,6 +293,9 @@ def pronounce_number(number, lang='', places=2, short_scale=True,
 
     Args:
         number: the number to pronounce
+        lang (str, optional): an optional BCP-47 language code, if omitted
+                              the default language will be used.
+        places (int): number of decimal places to express number to, default 2
```
ChanceNCounter, line 298: "number of decimal places to express" might be better. idk, i can't come up with good phrasing, except that the trailing 'to' is giving me fits

ChanceNCounter:
First pass looks good, marked unread for second pass after lunch.

Props and thanks for those type hints.

ChanceNCounter:
Well, I don't see any code changes (apart from the type hints) and neither does pytest, so, ✅

ChanceNCounter (approved):

lingua_franca/format.py
krisgesling: Good suggestion - fixed.

PR #197, opened 2021-05-28
Add configurable default timezone information

krisgesling:
#### Description
Adds configurable default timezone information for use by 

- Default tz-info can be set with `lingua_franca.time.set_default_tz`
- All datetime extractors default to `now_local` for `anchorDate`
- All localized functions now inject a tzinfo in all tz-naive datetime objects passed as arguments. Note tz-naive objects are presumed to be in UTC as there is no perfect answer here. 
  - Note this will also be done on methods that do not explicitly require tzinfo such as `nice_time()`
  - As such projects are encouraged to pass in timezone aware datetime objects where ever possible.
- Timezone info injection can be enabled/disable in lingua_franca.config (default: enabled) 
- Lingua Franca should always return a timezone aware datetime object in the configured default timezone

Extends and replaces PR #180  - squashed commits and fixed failing tests.

#### Type of PR
- [x] Bugfix
- [x] Feature implementation

#### Testing
Unit tests have been updated.

### DIFF
--- a/lingua_franca/internal.py
+++ b/lingua_franca/internal.py
@@ -467,12 +467,12 @@ def localized_function(run_own_code_on=[type(None)]):
                         kwargs[key] = to_local(value)
                 for idx, value in enumerate(args):
                     if isinstance(value, datetime) and value.tzinfo is None:
-                        args = args[:idx] + (to_local(value),) + args[idx + 1:]
+                        args = (*args[:idx], to_local(value), *args[idx + 1:])
 
             # Check if we're passing a lang as a kwarg
             if 'lang' in kwargs.keys():
                 lang_param = kwargs['lang']
-                if lang_param == None:
+                if lang_param is None:
                     warn(NoneLangWarning)
                     lang_code = get_default_lang()
                 else:
@@ -481,7 +481,7 @@ def localized_function(run_own_code_on=[type(None)]):
             # Check if we're passing a lang as a positional arg
             elif lang_param_index < len(args):
                 lang_param = args[lang_param_index]
-                if lang_param == None:
+                if lang_param is None:
                     warn(NoneLangWarning)
                     lang_code = get_default_lang()
                 elif lang_param in _SUPPORTED_LANGUAGES or \
--- a/lingua_franca/internal.py
+++ b/lingua_franca/internal.py
@@ -2,10 +2,12 @@ import os.path
 from functools import wraps
 from importlib import import_module
 from inspect import signature
-from sys import version
-from warnings import warn
 
+from warnings import warn
+from datetime import datetime
 from lingua_franca import config
+from lingua_franca.time import to_local
+
 
 _SUPPORTED_LANGUAGES = ("ca", "cs", "da", "de", "en", "es", "fr", "hu",
                         "it", "nl", "pl", "pt", "sl", "sv", "fa")
@@ -457,6 +459,15 @@ def localized_function(run_own_code_on=[type(None)]):
             func_params = list(func_signature.parameters)
             lang_param_index = func_params.index('lang')
             full_lang_code = None
+
+            # Check if we need to add timezone awareness to any datetime object
+            if config.inject_timezones:
+                for key, value in kwargs.items():
+                    if isinstance(value, datetime) and value.tzinfo is None:
+                        kwargs[key] = to_local(value)
+                for idx, value in enumerate(args):
+                    if isinstance(value, datetime) and value.tzinfo is None:
+                        args = args[:idx] + (to_local(value),) + args[idx + 1:]
 
             # Check if we're passing a lang as a kwarg
             if 'lang' in kwargs.keys():
--- a/lingua_franca/lang/parse_es.py
+++ b/lingua_franca/lang/parse_es.py
@@ -15,7 +15,8 @@
 #
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
-from dateutil.tz import gettz
+
+from lingua_franca.time import now_local
 from lingua_franca.lang.format_es import pronounce_number_es
 from lingua_franca.lang.parse_common import *
 from lingua_franca.lang.common_data_es import _ARTICLES_ES, _STRING_NUM_ES
@@ -367,7 +368,7 @@ def extract_datetime_es(text, anchorDate=None, default_time=None):
     if text == "":
         return None
     if anchorDate is None:
-        anchorDate = datetime.now()
+        anchorDate = now_local()
 
     found = False
     daySpecified = False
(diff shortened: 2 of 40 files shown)


ChanceNCounter:
Sorry, lost in notification spam. Tagging for review.

test/test_parse_es.py
```diff
@@ -132,36 +133,36 @@ def test_datetime_by_date_es(self):
         _now = datetime.now()
         relative_year = _now.year if (_now.month == 1 and _now.day < 11) else \
             (_now.year + 1)
-        self.assertEqual(extract_datetime_es("11 ene")[0],
+        self.assertEqual(extract_datetime_es("11 ene", anchorDate=_now)[0],
```
ChanceNCounter, line 136: Good catch o.o

ChanceNCounter:
I would like to clarify whether the Czech datetime extractor is injecting tzinfo in all the scenarios where @JarbasAl's logic does so in the other files.

lingua_franca/lang/parse_es.py
```diff
@@ -1020,17 +1021,18 @@ def date_found():
             datestr = datestr.replace(monthsShort[idx], en_month)
 
         temp = datetime.strptime(datestr, "%B %d")
-        temp = temp.replace(tzinfo=None)
+        if extractedDate.tzinfo:
+            temp = temp.replace(tzinfo=extractedDate.tzinfo)
+        # temp = to_local(temp)
```
ChanceNCounter, line 1026: stray

lingua_franca/lang/parse_fr.py
```diff
@@ -939,6 +943,9 @@ def date_found():
     if datestr != "":
         if not hasYear:
             temp = datetime.strptime(datestr, "%B %d")
+            if extractedDate.tzinfo:
+                temp = temp.replace(tzinfo=gettz("UTC"))
+                temp = temp.astimezone(extractedDate.tzinfo)
```
ChanceNCounter, line 948: What's going on here?

JarbasAl: Strptime does not have tzinfo, so its added automatically

JarbasAl: Otherwise dt comparison fails a few steps below

Tony763:
> I would like to clarify whether the Czech datetime extractor is injecting tzinfo in all the scenarios where @JarbasAl's logic does so in the other files.

Hi, @ChanceNCounter whan I did a Czech support, I took a English files and did 1:1 copy. If You need to change something, write what an I will look into it. 🙂

lingua_franca/internal.py
```diff
@@ -456,10 +458,19 @@ def _call_localized_function(func, *args, **kwargs):
...
             full_lang_code = None
 
+            # Check if we need to add timezone awareness to any datetime object
+            if config.inject_timezones:
+                for key, value in kwargs.items():
+                    if isinstance(value, datetime) and value.tzinfo is None:
+                        kwargs[key] = to_local(value)
+                for idx, value in enumerate(args):
+                    if isinstance(value, datetime) and value.tzinfo is None:
+                        args = args[:idx] + (to_local(value),) + args[idx + 1:]
```
ChanceNCounter, line 468: @krisgesling I am confused by this tuple's tupleyness

ChanceNCounter:
@Tony763 I see that now, thanks! I was just as confused by the snippet's absence from the English parser, but I've checked, and all paths add tzinfo to that temp var.

ChanceNCounter:
Apart from the remaining question on `internal.py`, I'd like to test that valid tzinfo isn't being overwritten.

lingua_franca/internal.py
krisgesling: Yeah I did a double take on it at first too. 
`('a')` provides a string; whilst
`('a',)` provides a single item tuple

I've switched it out to use iterable unpacking:
```
args = (*args[:idx], to_local(value), *args[idx + 1:])
```

hopefully this is more readable, though I also don't see it being used a lot so may still confuse some people.

test/test_format.py
```diff
@@ -402,15 +403,16 @@ def setUpClass(cls):
                     cls.test_config[sub_dir.parts[-1]] = json.loads(f.read())
 
     def test_convert_times(self):
-        dt = datetime.datetime(2017, 1, 31,
-                               13, 22, 3)
+        dt = datetime.datetime(2017, 1, 31, 
+                               13, 22, 3, tzinfo=default_timezone())
```
JarbasAl, line 407: small note, i think its good to be explicit here, but passing this is fully optional, part of what this PR does is handle this transparently

the tests only need tzinfo added in returned results, all input dates will get it appended internally

in a sense this last commit adding all those tzinfos in optional places is now removing the testing for the functionality added by this PR

as i said i think it's good to be explicit in what is being tested, this comment is merely informative, not a request for a change

krisgesling: Yeah it is an issue I think. 

With the naive datetimes having UTC tzinfo added, from memory the tests will fail if that's not your configured default unless you explicitly pass an aware object. So it is a bit of a patch job - we really need more detailed testing that checks those different scenarios.

krisgesling:
I've just added an attempted test for using an explicitly different timezone. However it also suggests that one of the dot points from the PR description is false:
> Lingua Franca should always return a timezone aware datetime object in the configured default timezone

This is also what the docstring says, but it possibly should be:
> Lingua Franca should always return a timezone aware datetime object. If the method accepts a datetime object as an argument, and a timezone aware object is passed, the return value will be in the same timezone. If there is no datetime object argument, or a timezone naive object is passed, the configured default timezone will be used.

lingua_franca/internal.py
ChanceNCounter: It's significantly more readable, in that I thought the outer type was a list, but now I'm wondering why `args` is a tuple 👅

ChanceNCounter:
I agree, and had been reading,

> always return a timezone aware datetime object in the configured default timezone [if LF had to add the tzinfo itself]

We definitely don't want to clobber tz-aware datetimes *back* to the configured tz!

ChanceNCounter (approved):
Thought I pressed approve on my previous comment. Whoops =)

ChanceNCounter (approved):

krisgesling:
What about we just remove the word "user's":
> [:obj:`datetime`, :obj:`str`]: 'datetime' is the extracted date
>             as a datetime object in the user's local timezone.

becomes
> [:obj:`datetime`, :obj:`str`]: 'datetime' is the extracted date
>             as a datetime object in the local timezone.

ChanceNCounter:
That'd work.

PR #200, opened 2021-06-10
Clarify docstring that valid tz won't be mutated

krisgesling:
#### Description
Small clarification - returned data will be in the "local timezone" which will be either the timezone of the datetime object passed into the method OR the configured default timezone. With the word "user" included it sounded like the return value would always be in the user's configured default timezone.

#### Type of PR
- [x] Documentation improvements


ChanceNCounter (approved):

ChanceNCounter:
Happy #200, Lingua Franca!

PR #201, opened 2021-06-11
Implement function to tie percent sign to number

### LINKED ISSUE #196: Normalizer mishandles "X%.", returns "X %."
`normalize("Set Volume to 50%.") -> "Set Volume to 50 %."`

This is bad. It should probably, at worst, return `"Set Volume to 50 % ."`

Badboy-16:
#### Description
Fixes #196 

Adds a function to tie the percent sign (`%`) to the preceding numbers.

#### Type of PR

- [x] Bugfix
- [ ] Feature implementation
- [ ] Test improvements

#### Testing
Testing added in `test_parse_common.py`. (`test_normalize()`)

### DIFF
--- a/lingua_franca/lang/parse_common.py
+++ b/lingua_franca/lang/parse_common.py
@@ -162,6 +162,28 @@ class Normalizer:
         utterance = " ".join(words)
         return utterance
 
+    @staticmethod
+    def _is_number(string):
+        try:
+            number = float(string)
+        except:
+            return False
+        return True
+
+    def tying_percentage(self, utterance):
+        words = self.tokenize(utterance)
+        output = []
+        for idx, w in enumerate(words):
+            if w.startswith('%') and idx > 0:
+                if self._is_number(output[-1]) is False:
+                    output.append(w)
+                else:
+                    output[-1] += w
+            else:
+                output.append(w)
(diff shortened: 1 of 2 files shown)


ChanceNCounter:
Sorry, I think I failed to communicate the core problem with #196.

The larger issue is that the normalizer is returning `"X %."`, instead of `"X % .", because it doesn't generally insert a space before punctuation. Consumers looking for that percent sign in isolation miss it when it has punctuation attached to it.

The space between the number and the percent sign is apparently useful downstream, so we might have to revert that part.

Badboy-16:
It's ok @ChanceNCounter 

So in that case, should the behaviour of normalizer be that it inserts a space between the percent sign and the punctuation? If so, I think I can revert my previous code and add a function to check if any percent signs preceding any punctuations, and if yes, insert a space between them.

ChanceNCounter:
That's the desired behavior, yes. I dunno if the other discussion is closed forever =P but that's the part that should change for now. The trouble with changing behavior such as the separate percent sign is that existing code might be looking for that output.

We can make changes like that, with an appropriate deprecation period, but it's no small thing and the discussion has to brew for a while. Apparently our predecessors put that space in there on purpose.

PR #206, opened 2021-07-30
Changes for Basque language

e-gor:
Changes for Basque language
===========================

#### Description

These changes, along with those sent to other repos, make Mycroft work in Basque language.

Made by Elhuyar, Talaios and Skura

#### Type of PR

Feature implementation


devs-mycroft:
Hello, @e-gor, thank you for helping with the Mycroft project! We welcome everyone
into the community and greatly appreciate your help as we work to build an AI
for Everyone.

To protect yourself, the project, and users of Mycroft technologies we require
a Contributor Licensing Agreement (CLA) before accepting any code
contribution. This agreement makes it crystal clear that along with your
code you are offering a license to use it within the confines of this project.
You retain ownership of the code, this is just a license.

Please visit https://mycroft.ai/cla to initiate this one-time signing. Thank
you!

test/test_parse_eu.py
```diff
@@ -0,0 +1,231 @@
...
+                         "20 30 41")
+        self.assertEqual(normalize("hogeita hamabi zaldi", lang="eu"),
+                         "32 zaldi")
+        self.assertEqual(normalize("ehun zaldi", lang="eu"),
+                         "100 zaldi")
+        self.assertEqual(normalize("ehun eta hamaika zaldi", lang="eu"),
+                         "111 zaldi")
+        self.assertEqual(normalize("laurehun eta bat behi zeuï¿½en",
+                                   lang="eu"),
+                         "401 behi zeuï¿½en")
```
ChanceNCounter, line 64: I see a few cases with what appear to be the wrong unicode characters. Is this a font thing?

e-gor: There are occurrences of that character in the portuguese and spanish files, and I think the reason for their being in the basque files is because they were taken as a base. I have removed and corrected them from the basque files and also from the others, I am pretty confident they were there due to wrong encoding from copy/pasting and not for correcting input in another encoding (it's just in a couple of cases from many possible ones).

ChanceNCounter: Welcome to the glorious world of special characters in code ;) kicked back up near the top of my code review list, thanks for addressing it quickly

lingua_franca/format.py
```diff
@@ -566,3 +567,53 @@ def nice_response(text, lang=''):
         assertEqual(nice_response_de("10 ^ 2"),
                          "10 hoch 2")
     """
+
+@localized_function(run_own_code_on=[FunctionNotLocalizedError])
+@localized_function()
+def nice_relative_time(when, relative_to=None, lang=None):
```
krisgesling, line 573: If we're adding in a new function, we really need unit tests for it.

It would really be better as a separate PR since it goes beyond the scope of adding Basque.

lingua_franca/res/text/fa-ir/normalize.json
```diff
@@ -0,0 +1,141 @@
+{
```
krisgesling, line 1: This file got deleted, so I think it's just come along for the ride and you haven't intended for it to be here.

krisgesling (changes requested):
Hey Igor, 

Thanks for nudge, we've been very heads down on the Mark II and this one fell off my radar. 

It looks great though. Just two items that need a look at below and then we can merge it in.

Each release we title with "Hello World" in one of the added languages. For our next release, it would be great to know how you say "Hello World" in Basque?

Thanks

test/test_parse_eu.py
```diff
@@ -0,0 +1,231 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from datetime import datetime
+import unittest
+
+from lingua_franca.parse import (normalize, extract_numbers, extract_number,
+                                 extract_datetime, extract_datetime_eu,
+                                 isFractional_eu)
```
krisgesling, line 21: ah these, eu specific functions should get imported directly `from lingua_franca.lang.parse_eu`

Though if the language is set you _should_ be able to just call `extract_datetime()`

e-gor:
Thanks very much for the feedback! We will fix the issues as soon as possible.

e-gor:
By the way,

> Each release we title with "Hello World" in one of the added languages. For our next release, it would be great to know how you say "Hello World" in Basque?

In Basque we would say "Kaixo, mundua!".

lingua_franca/res/text/fa-ir/normalize.json
e-gor: Done.

test/test_parse_eu.py
e-gor: Fixed for extract_datetime_eus.

For isFractional_eu, the problem seems to be that the function only exists for some languages (cs, pl and now eu) and it is not in parse.py. The problem has not arisen before because the other two languages have no tests for it... So I think I shall leave this as it is.

lingua_franca/format.py
e-gor: Test added and made separate PR for this (https://github.com/MycroftAI/lingua-franca/pull/231).

Now Basque changes PR only contains Basque language stuff.

The nice_relative_time PR should be merged before the Basque one, since the Basque one localizes also the nice_relative_time function and this must therefore exist.

e-gor:
All requested changes done, @krisgesling.

- nice_relative_time function moved to a new PR, with a test included.
- Removed the file that had been deleted but we included in the PR.
- Fixed the extract_datetime_eu issue in the Basque PR.

So now we have a PR with the nice_relative_time function (https://github.com/MycroftAI/lingua-franca/pull/231) and another one with the Basque language functions (https://github.com/MycroftAI/lingua-franca/pull/206). The nice_relative_time PR should be merged before the Basque one, since the Basque one localizes also the nice_relative_time function and this must therefore exist.

test/test_parse_eu.py
```diff
@@ -0,0 +1,228 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from datetime import datetime
+import unittest
+
+from lingua_franca.parse import (normalize, extract_numbers, extract_number,
+                                 extract_datetime, isFractional_eu)
```
krisgesling, line 20: Hey, the problem is that this function definition does not exist in the `lingua_franca.parse` namespace so it can't import `isFractional_eu` from this module. You can see that from the test output here: 
https://github.com/MycroftAI/lingua-franca/runs/7036546990?check_suite_focus=true

You would need to have:
```
from lingua_franca.parse import (normalize, extract_numbers, extract_number,
                                 extract_datetime)
from lingua_franca.lang.parse_eu import isFractional_eu
```

Or we'd need to register `is_fractional` as a top-level localizable function, but it's not really intended as a public function. At least I haven't seen it being used as such.

We should probably take the opportunity to fix up the casing of the function name too. It should really be `is_fractional_eu`. But I know this was copied from other languages.

e-gor: Done!

Also fixed some failing tests (which were not happening when the PR was made almost a year ago; I guess they were caused by some errors in the merging of the updates happened since).

krisgesling (approved):

PR #212, opened 2021-10-18
Implement Azerbaijani support

saymantech-co:
Implement Azerbaijani support(az_AZ)

Type of PR
- [X] Feature implementation

‌Testing
- pytest test/test_format_az.py
- pytest test/test_parse_az.py


siathon:
For the date time stuff to work, the "az_AZ" locale needs to be generated in the OS, that's why the tests failed. Is this acceptable or should I find another way to handle the date time stuff?

krisgesling:
Sorry I missed the notification for your last message.

My first reaction is that we should expect the distribution packaging this to ensure the correct locales are available, and yes we can add that to the CI job here too.

I'm not at all familiar with the locale module so not sure what side effects we might be creating by switching the locale back and forth. There are also warnings of it not being threadsafe. 

Would it be sufficient for us to install `language-pack-az`?

krisgesling:
#216 looks like it generated the locale without issue, so we can try it but I'm very interested to hear from anyone who's worked with locale before. Or if there are other options we can go with.

lingua_franca/lang/common_data_az.py
```diff
@@ -0,0 +1,297 @@
...
+        if is_last:
+            return "ncu"
+        return "uncu"
+
+    if last_vowel == ["ö", "ü"]:
+        if is_last:
+            return "ncü"
+        return "üncü"
+
+def _get_full_time_ak(hour):
```
krisgesling, line 228: what does "ak" refer to in these function names? I've seen it throughout this PR

lingua_franca/res/text/az-az/date_time.json
```diff
@@ -0,0 +1,130 @@
...
+      "date_time": "{formatted_date} {formatted_time}"
+  },
+    "weekday": {
+    "0": "bazar ertəsi",
+    "1": "çərşənbə axşamı",
+    "2": "çərşənbə",
+    "3": "cümə axşamı",
+    "4": "cümə",
+    "5": "şənbə",
+    "6": "bazar"
```
krisgesling, line 51: I'm guessing "bazar" is used in informal speech rather than "bazar günü"?

test/test_format_az.py
```diff
@@ -0,0 +1,567 @@
...
+        # value is platform dependent so better not use in tests?
+        # self.assertEqual(
+        #    pronounce_number(sys.float_info.min), "iki nöqtə iki iki times "
+        #                                          "on to the power of "
+        #                                          "mənfi üç yüz "
+        #                                          "və səkkiz")
+        # self.assertEqual(
+        #    pronounce_number(sys.float_info.max), "bir nöqtə yeddi doqquz "
+        #                                          "times on to the power of"
+        #                                          " üç yüz və səkkiz")
```
krisgesling, line 230: I'd agree with the comment - this can be deleted unless you see specific value in it?

test/test_format_az.py
```diff
@@ -0,0 +1,567 @@
...
+            "yüz trilyon, otuz dörd min milyard, "
+            "iki yüz doxsan doqquz milyon, yeddi yüz "
+            "doxsan iki min, dörd yüz əlli səkkiz")
+        self.assertEqual(
+            pronounce_number(10000000000, short_scale=True),
+            "on milyard")
+        self.assertEqual(
+            pronounce_number(1000000000000, short_scale=True),
+            "bir trilyon")
+        # TODO maybe beautify this
```
krisgesling, line 258: Is this relevant to the AZ test or just cut and pasted from the other file?

test/test_format_az.py
```diff
@@ -0,0 +1,567 @@
...
+                                          short_scale=False),
+                         "on səkkiz milyardıncı")
+        self.assertEqual(pronounce_number(18e12, ordinals=True),
+                         "on səkkiz trilyonuncu")
+        self.assertEqual(pronounce_number(18e18, ordinals=True,
+                                          short_scale=False), "on səkkiz "
+                                                              "trilyonuncu")
+
+# def nice_time(dt, lang="en-us", speech=True, use_24hour=False,
+#              use_ampm=False):
```
krisgesling, line 318: It looks like this has been copied across multiple languages :laughing: 

I'll remove it from all.

test/test_parse_az.py
```diff
@@ -0,0 +1,429 @@
+#
+# Copyright 2017 Mycroft AI Inc.
```
krisgesling, line 2: Would be good to use the current year in each of these license headers.

krisgesling (commented):
Hey sorry it's taken a while to get to this. 

It's always a challenge reviewing extensive code in another language! But it looks pretty great.

There are a few little questions below but nothing of concern. The only thing is whether we really need to set the locale or not.

siathon:
> Sorry I missed the notification for your last message.
> 
> My first reaction is that we should expect the distribution packaging this to ensure the correct locales are available, and yes we can add that to the CI job here too.
> 
> I'm not at all familiar with the locale module so not sure what side effects we might be creating by switching the locale back and forth. There are also warnings of it not being threadsafe.
> 
> Would it be sufficient for us to install `language-pack-az`?

The reason switching the locale back and forth was when running pytest, after running AZ tests, because of changing the locale in AZ tests, all other language tests failed, So in AZ, I returned the locale to the previous locale.

if that is not necessary and only passing test_parse_az and test_format_az is enough, the switching back and forth is not required and can be removed.

lingua_franca/lang/common_data_az.py
siathon: 1) In Azerbaijani language instead of prepositions we have "ək"(I didn't want to use special characters in function name so i used "ak"). and the "ək" is added to the end of word. 
2) In Azerbaijani  we have two groups of vowels, and the "ək" used in word differs according to words last vowel and the ending of the word.

I'll explain with an example.
In AZ "five" is "beş" and "six" is "altı".
when saying time, in english both is same: at five, at six
but in Az "at five" is "beşdə"(the "ək" is "də") but "at six" is "altıda"(the "ək" is "da").

I use these functions to generate the correct ək for the words.

lingua_franca/res/text/az-az/date_time.json
siathon: No, the equivalent for sunday is "bazar", "bazar günü" means "on sunday" and is the same other days, "şənbə günu" is "on saturday",...

test/test_format_az.py
siathon: No, sorry I used english files as template and edited them, and seems i have forgotten to remove some of non relevant comments.

test/test_format_az.py
siathon: yeah, most of these comments are the result of using the english files as template. I'll review and remove them.

test/test_parse_az.py
siathon: Ok, I will fix them.

lingua_franca/res/text/az-az/date_time.json
krisgesling: and this is why we don't use automated translation tools for this stuff :smile:

lingua_franca/lang/common_data_az.py
krisgesling: Cool, thanks for the explanation - I'm always interested in seeing how different languages do things. Hopefully helps us to identify assumptions we make about the structure of all languages.

krisgesling:
With the AZ language pack installed and all the locale setting removed from `parse_az.py`. The following 4 test cases fail:
```python
        testExtract("3 avqustda anama zəng etməyi xatırlat",
                    "2017-08-03 00:00:00", "anama zəng etməyi xatırlat")
        testExtract("iyulun 4 də atəşfəşanlıq al",
                    "2017-07-04 00:00:00", "atəşfəşanlıq al")
        testExtract("dekabr 3",
                    "2017-12-03 00:00:00", "")
        testExtract("5 iyun 2017 ci il axşamı anama zəng etməyi xatırlat",
                    "2017-06-05 19:00:00", "anama zəng etməyi xatırlat")
```

Is that what you're seeing also?

Are these the only test cases that would rely on the locale getting set or does it point to something else going on?

siathon:
> With the AZ language pack installed and all the locale setting removed from `parse_az.py`. The following 4 test cases fail:
> 
> ```python
>         testExtract("3 avqustda anama zəng etməyi xatırlat",
>                     "2017-08-03 00:00:00", "anama zəng etməyi xatırlat")
>         testExtract("iyulun 4 də atəşfəşanlıq al",
>                     "2017-07-04 00:00:00", "atəşfəşanlıq al")
>         testExtract("dekabr 3",
>                     "2017-12-03 00:00:00", "")
>         testExtract("5 iyun 2017 ci il axşamı anama zəng etməyi xatırlat",
>                     "2017-06-05 19:00:00", "anama zəng etməyi xatırlat")
> ```
> 
> Is that what you're seeing also?
> 
> Are these the only test cases that would rely on the locale getting set or does it point to something else going on?

yeah, These are the cases that rely on locale being set.

siathon:
@krisgesling is using the locale setting alright or should I try another way to handle these cases?

krisgesling:
I'd really like it if we could find another way. But I don't have a suggestion for what that is sorry.

siathon:
Done, no locale setting required for parsing datetime.
575299c09aae85844d145e59d5e311b9bf6a4a5a

Issue #203, opened 2021-06-15
The test_format_ and test_parse_fa tests has errors


forslund:
When running the test suite I get a couple of errors like this:
```
        def testExtract(text, expected_date, expected_leftover):
            res = extractWithFormat(normalize(text))
            self.assertEqual(res[0], expected_date, "for=" + text)
            self.assertEqual(res[1], expected_leftover, "for=" + text)

        testExtract("الان ساعت اینه",
>                   "2017-06-27 13:04:00", "ساعت اینه")

test/test_parse_fa.py:116:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test/test_parse_fa.py:112: in testExtract
    self.assertEqual(res[0], expected_date, "for=" + text)
E   AssertionError: '2017-06-27 15:04:00' != '2017-06-27 13:04:00'
E   - 2017-06-27 15:04:00
E   ?             ^
E   + 2017-06-27 13:04:00
E   ?             ^
E    : for=الان ساعت اینه

```

Best guess, there is something with the timezone?

ChanceNCounter:
Yep. TZ PR broke it. I wonder why the tests passed on GitHub. It's not just Farsi, either.

ChanceNCounter:
Update: Jarbas and I are on it. At least two of these are due to a loss in translation during PR review. Others were masked by naming errors in the Farsi PR.

Still others, which aren't occurring on @forslund's computer but are on mine, remain to be investigated (including one in Swedish oh boy) but I am *assuming* we introduced the bug at the same time

Further info to follow.

ChanceNCounter:
Working in spurts, I have yet to determine why the default time zone seems to be changing mid-run. However, many of the datetime tests specify the default time zone when they create their reference datetime objects. I'm now trying to determine if the bug is in the code, or in the tests; is something else interfering with the mechanism that ensures the right time zone, or is it simply that pytest doesn't run tests in order, and timing becomes a factor?

JarbasAl:
https://github.com/HelloChatterbox/lingua-nostra/pull/32

PR #215, opened 2021-11-10
fix potential circular import

ChanceNCounter:
Fixes potential circular import during the import-and-crawl phrase of localized function caching, where a localized file attempts to reference its parent. By initializing the rest of the parent file *before* attempting to populate the localized function dictionary, it seems to delay the whole thing (and therefore the import-and-crawl phrase) long enough to avoid the ImportError.

In reality, when the potential error hits, the parent file is still partially initialized. In other words, this could technically still be described as a circular import, but, because Python is Python, does not have the effect of one; whatever the localized file is attempting to reference in its parent, that function will necessarily already have been parsed before the localized function is imported.

This is known to work, because HelloChatterbox ran into the problem, and I've already tested the fix against their code.

### DIFF
--- a/lingua_franca/format.py
+++ b/lingua_franca/format.py
@@ -37,9 +37,6 @@ _REGISTERED_FUNCTIONS = ("nice_number",
                          "nice_response",
                          "nice_duration")
 
-populate_localized_function_dict("format", langs=get_active_langs())
-
-
 def _translate_word(name, lang=''):
     """ Helper to get word translations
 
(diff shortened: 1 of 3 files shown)


ChanceNCounter:
Note: this bug doesn't exist yet, but now it's impossible to write!

Additional note: Local unit tests are currently wonky because of #203, pending #209. However, there is no change in test status, either on GitHub or On My Computer. Moot point if #209 gets in before this does.

PR #226, opened 2022-02-28
Fix #224 - parsing 'in 2007'

### LINKED ISSUE #224: "In 2007" should be parsed as year
**Describe the bug**
I know for Mycroft it reasonable to parse 2007 as 20:07. But this is a separate library, and for 2007 should be a year.

**To Reproduce**
```python
from lingua_franca.parse import *
lingua_franca.load_language('en')
print(lingua_franca.parse.extract_datetime('in 2007'))
[datetime.datetime(2022, 2, 21, 20, 7), '']
```

**Expected behavior**
[datetime.datetime(2007, 0, 0, 0, 0), '']

**env**
lingua_franca-0.4.2-py3-none-any.whl
python3.9

krisgesling:
#### Description
Fixes #224 - extract_datetime parsing of phrases like "in 2007"

Whilst this isn't ideal behaviour, it is better than returning a time.

This is a port of https://github.com/OpenVoiceOS/ovos-lingua-franca/pull/9

#### Type of PR
- [x] Bugfix

#### Testing
Test added

### DIFF
--- a/lingua_franca/lang/parse_en.py
+++ b/lingua_franca/lang/parse_en.py
@@ -694,7 +694,8 @@ def extract_datetime_en(text, anchorDate=None, default_time=None):
     timeQualifiersAM = ['morning']
     timeQualifiersPM = ['afternoon', 'evening', 'night', 'tonight']
     timeQualifiersList = set(timeQualifiersAM + timeQualifiersPM)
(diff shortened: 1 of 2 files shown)


forslund (approved):
Looks good! and seems to work as expected!

PR #237, opened 2022-08-02
Add more BCP-47 default language codes

NeonDaniel:
#### Description
Add more default BCP-47 languages based on https://cloud.google.com/speech-to-text/docs/languages

Also defined: 'ga-ie'

#### Type of PR
If your PR fits more than one category, there is a high chance you should submit more than one PR. Please consider this carefully before opening the PR.
_Either delete those that do not apply, or add an x between the square brackets like so: `- [x]`_
- [ ] Bugfix
- [X] Feature implementation
- [ ] Test improvements

#### Testing
`get_full_lang_code` will now resolve more BCP-47 language codes

#### Documentation
No functional changes.

#### CLA
To protect you, the project, and those who choose to use Mycroft technologies in systems they build, we ask all contributors to sign a Contributor License Agreement.

This agreement clarifies that you are granting a license to the Mycroft Project to freely use your work.  Additionally, it establishes that you retain the ownership of your contributed code and intellectual property.  As the owner, you are free to use your code in other work, obtain patents, or do anything else you choose with it.

If you haven't already signed the agreement and been added to our public [Contributors repo](https://github.com/mycroftAI/contributors) then please head to https://mycroft.ai/cla to initiate the signing process.


krisgesling:
Hey Daniel, 

This won't break anything, but I'm curious what purpose it serves to have a big list of language codes?

The one thing that concerns me is that these have been sourced from the Google Cloud site and I don't expect that they have positive copyright provisions covering that content. Even though they're part of the BCP-47 spec, it still falls into a legal grey area:
https://mycroft.ai/blog/licensing-a-language-how-the-copyright-system-abuses-fundamental-human-rights/

Longer term we should probably move to the current standard ISO 639.2 but that's clearly well outside the scope of this change.

NeonDaniel:
> This won't break anything, but I'm curious what purpose it serves to have a big list of language codes?

The `get_full_lang_code` is generically useful for getting a default BCP-47 code from an ISO 639 code since those are what are used in core for resources. I use the method in a few places to make sure `Message` objects always get the `BCP-47` code if there was some language translation involved.

> The one thing that concerns me is that these have been sourced from the Google Cloud site and I don't expect that they have positive copyright provisions covering that content. Even though they're part of the BCP-47 spec, it still falls into a legal grey area: https://mycroft.ai/blog/licensing-a-language-how-the-copyright-system-abuses-fundamental-human-rights/

Fair enough. This was just a source I had referenced in code previously, but I'll find the upstream spec from ISO or some open source reference.

> Longer term we should probably move to the current standard ISO 639.2 but that's clearly well outside the scope of this change.

I agree on both points. There is hopefully already some implementation we could use to translate between the 2 specs but that would be a bigger transition to plan.

lingua_franca/internal.py
```diff
@@ -18,24 +18,79 @@
                                  "fa-ir", "pt-pt", "ru-ru", "sl-si",
                                  "sv-se", "tr-tr", "eu-eu")
```
NeonKirill, line 20: please use `bidict` here as sometimes we need a key by value

NeonDaniel: By definition, we should always get the key as `val.split('-', 1)[0]` in this list

